PR #20, opened 2016-03-21
reorganize and update all dependencies

csamsel:
compile tested


goekay:
such extraction of all versions into properties seems excessive and pollutes the already busy looking pom file. moreover, it disconnects the maintainer from the context in which a dependency is used and therefore might cause batch-updating every version without any thinking. i believe we should define version properties of dependencies that are 1) used more than once and/or 2) have a significant importance to the project.

PR #36, opened 2018-03-12
Update CentralSystemService15_Service.java

JavaIsJavaScript:
Even if it's a dummy implementation if this gets called by a Charge Point this will return an "OccurenceConstraintViolation".
According to the OCPP specifications the "DataTransfer.conf" is REQUIRED to return a DataTransferStatus.


goekay:
good catch, thanks for your contribution!

JavaIsJavaScript:
Working on SteVe myself, so i'll contribute where i can :)

PR #37, opened 2018-03-13
Fix UpdateFirmware

JavaIsJavaScript:
Was able to call Update Firmware without the retrieveDate parameter.

Absence of the retrieveDate paramater will cause an "OccurenceConstraintViolation". As described in the OCPP specification, this parameter is mandatory.


goekay:
once again, good catch. thanks!

PR #40, opened 2018-03-14
Add 1.5 features for 1.6

JavaIsJavaScript:


JavaIsJavaScript:
Forgot to change getGetConf on line 196 in Ocpp16Controller, still has ConfigurationKeyEnum.OCPP_15_MAP, need to change this to ConfigurationKeyEnum.OCPP_16_MAP

goekay:
i guess you want to help with implementation of the 1.6 client, which i appreciate, but i cannot accept this PR because of two reasons:
- it contains too much changes and commits that are already applied in the ocpp-1.6-impl branch. i guess this is due to you forking the repo with the old structure. the PR should be atomar in the sense hat it should contain only the commits/changes related to the case.
- i don't want to blindly copy all 1.5 classes and jsps, and rename them to 1.6. this will be a maintenance hell in the future. as we did with the 1.6 service impl, i want to unify/generalize the common parts in the code base first.

JavaIsJavaScript:
I get that, one change might be still handy which is the ConfigurationKeyEnum with all the new ConfigurationKeys in 1.6

PR #43, opened 2018-03-19
Implement TriggerMessage Function

JavaIsJavaScript:


goekay:
thanks for the implementation. please modify the PR to contain 2 fixes:
- the java code should use the same style as with the rest of the code base
- trigger message can be used for _multiple_ stations, since connectorId is optional. there is no connector requirement to ask all your stations for a hearbeat.

JavaIsJavaScript:
Quick question,

In the OCPP specification at SetChargingProfile.req it says:  If connectorId = 0,
the message contains an overall limit for the Charge Point.

and at GetCompositeSchedule.req it says: When connectorId=0, the Charge Point will calculate
the expected consumption for the grid connection

even though the connectorId is required at both are these also usable on multiple stations at once?

goekay:
this question is not related to the pull request, let's discuss this in a new issue: #44

PR #51, opened 2018-03-27
Remove deprecated ConfigurationKeys for 1.6

JavaIsJavaScript:


goekay:
we had multiple issues and commits regarding the ConfigurationKeyEnum. @JavaIsJavaScript since you started the the changes w.r.t. ocpp 1.6, can you please make sure that the state after this commit is stable? with this i mean proof-reading while comparing the impl to the spec to make sure that there are no mistakes. unfortunately, i do not have the time for it. 

until i get a feedback from you, i will wait to merge this PR.

JavaIsJavaScript:
I have checked all the ConfigurationKeys one by one and everything should be correct now, The 3 that got changed were being maintained for compatibility for earlier releases. My fault that I thought that all old ConfigurationKeys would be used in 1.6. All that would need to be done now is implement the "Smart Charging" ConfigurationKeys if or when the "Smart Charging" profile gets implemented

PR #65, opened 2018-05-15
fix empty conf. key list

JavaIsJavaScript:
Pressing "Perform" without selecting a Charge Point in ChangeConfiguration makes the conf. key list empty.

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/web/controller/Ocpp16Controller.java
+++ b/src/main/java/de/rwth/idsg/steve/web/controller/Ocpp16Controller.java
@@ -116,6 +116,17 @@ public class Ocpp16Controller extends Ocpp15Controller {
         }
         return REDIRECT_TASKS_PATH + getClient15().getConfiguration(params);
     }
+    
+    @RequestMapping(value = CHANGE_CONF_PATH, method = RequestMethod.POST)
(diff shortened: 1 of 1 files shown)


goekay:
help me if i understand it correctly: if you press perform without selecting a charge point, you get a validation error page. you want to preserve the selected configuration key on that page to make it easier to continue?

JavaIsJavaScript:
All the configuration keys dissapear when you don't select a Charge Point and press perform

goekay:
hmm. this problem exists in older ocpp version controllers, too. so i would look for a general solution.

JavaIsJavaScript:
not sure if this happens in older ocpp version, will test it now, maybe it's because in ocpp1.6 conf. keys can be ReadOnly or ReadWrite

goekay:
it for sure exists in ocpp 1.2. just tested it. the line `model.addAttribute("ocppConfKeys", getConfigurationKeys());` is missing in postChangeConf there, too.

JavaIsJavaScript:
maybe handle this problem like GetConfiguration is handled and add "postChangeConf" to all the controllers since all ocpp versions need a different list of configuration keys?

goekay:
i have implemented a solution that modifies getConfigurationKeys. please tell me if you approve it. i am committing it in a few minutes.

JavaIsJavaScript:
works like a charm :)

goekay:
great. thanks for your help and involvement!

PR #151, opened 2019-06-27
Implemented optional email notifications 

abramowitzK:
This was useful for my organization when monitoring test chargers. We would really like to implement a notification for UpdateFirmware requests too but in my limited time I was not able to figure this out :).

Anyway figured I might as well contribute this back in case it is useful to anyone else.


goekay:
thanks for your contribution. before i can merge the pull request, i have two minor modification requests:
- the new NotificationFeatures contain `JSON` in the text, which is not correct. as you implemented it, this feature applies to all (JSON and SOAP) stations. please remove it.
- you are making a `null` check with `transactionId` (which is the right thing to do!), but in the implementation it actually can never be `null` and therefore can be considered as an `int`. the null check is not necessary. (this must be a relict of some previous implementation i cannot remember. thanks for making me aware of it. i will change the return type of `OcppServerRepository.insertTransaction` from `Integer` to `int` as soon as possible after merging.)

abramowitzK:
I've made the changes you requested. Thanks for the project!

Issue #142, opened 2019-06-11
Add Docker support


aakobordenau:
Hi, is there any interest in adding docker support to steve?

goekay:
yes. there is even an initial version of dockerized steve. but since i am not an expert in this area, i do not know the best practices and therefore i was hesitant to check it in. because of the same reasons, i do not think that i can maintain the docker version _at the moment_. i simply need time to delve into the world of docker which i lack right now.

aakobordenau:
Nice! Can I take a look at the draft anywhere?
I started to dockerize steve as well, but I may miss something because I didn't really dive deep enough into the project yet.

goekay:
sure. actually, my colleague @muued worked on "dockerization" of steve. and his changes can be found in the docker branch of [his fork](https://github.com/muued/steve). this is the relevant commit: https://github.com/muued/steve/commit/6211b9ef620cb961efd353041eb3463aa4ed2b66

lingster:
@aakobordenau @goekay I've just submitted a PR for this request. Could someone kindly review and provide feedback? You should be able to start Steve out of the box now!

PR #163, opened 2019-08-14
Add docker profile to fix issue #142

lingster:
Added Dockerfile and docker-compose.yml file

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,6 @@
+# mariadb setup 
+MYSQL_ROOT_PASSWORD=changeme
+#MYSQL_ROOT_HOST=, 
+MYSQL_DATABASE=stevedb
+MYSQL_USER=steve
+MYSQL_PASSWORD=changeme
--- a/.env
+++ b/.env
@@ -1,6 +1,0 @@
-# mariadb setup 
-MYSQL_ROOT_PASSWORD=changeme
-#MYSQL_ROOT_HOST=, 
-MYSQL_DATABASE=stevedb
-MYSQL_USER=steve
-MYSQL_PASSWORD=changeme
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN wget --no-verbose https://github.com/jwilder/dockerize/releases/download/$DO
     && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
 
 
-EXPOSE 8080
+EXPOSE 8180
 EXPOSE 8443
 WORKDIR /code
 
@@ -23,7 +23,7 @@ COPY . /code
 
 # Wait for the db to startup(via dockerize), then 
 # Build and run steve, requires a db to be available on port 3306
-CMD dockerize -wait tcp://db:3306 -timeout 60s && \
+CMD dockerize -wait tcp://mariadb:3306 -timeout 60s && \
 	mvn clean package -Pdocker -Djdk.tls.client.protocols="TLSv1,TLSv1.1,TLSv1.2" && \
 	java -jar target/steve.jar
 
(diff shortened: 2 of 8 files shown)


csamsel:
We cant have something like adminer in there. The MariaDB image would also require a tag.

lingster:
I've fixed the version of maria db and removed adminer (although the docker-compose example came from  here: https://hub.docker.com/_/mariadb ) and updated the startup script accordingly

csamsel:
Sorry for being picky but i'd suggest using a tag of a major release of MariaDB. e.g. 10.4
that way we would get security fixes but not major changes potentially breaking something. Else that looks fine for me. Havent had the chance to test it though.

goekay:
thanks for your work! i kind of find it not nice to pollute the repository this much with many scattered files for docker support. 
- is it possible to pull the contents of `.env` into `docker-compose.yml`?
- is it possible to pull the commands in `startup.sh` into `Dockerfile`?

i find the solution referenced in [my comment](https://github.com/RWTH-i5-IDSG/steve/issues/142#issuecomment-503620485) a little bit better since it is more compact. and therefore i would like to ask @lingster the following: what did you see in that solution as problematic and wanted to improve?

lingster:
@goekay I've moved the .env into the docker-compose file. For the startup.sh, the PR I have put together uses dockerize which specifically will wait until the mariadb has started and is ready to accept connections. This is required so that mvn package can run successfully. 

As I'm not so experienced with this project, which is the more correct way to build and run the project:
`mvn package -P<profile> `
or
`mvn -B -V -e clean package -DskipTests -Djdk.tls.client.protocols="TLSv1,TLSv1.1,TLSv1.2" && java -jar target/steve.jar`

For the [solution](https://github.com/muued/steve/commit/6211b9ef620cb961efd353041eb3463aa4ed2b66) although it is indeed more concise, the Dockerfile modifies the main.properties via a sed script which could be prone to breaking and is less flexible. Having a specific "docker" profile seems to make more sense, given there are already dev and test profiles too. 

@csamsel : mariadb version amended as requested

> thanks for your work! i kind of find it not nice to pollute the repository this much with many scattered files for docker support.
> 
>     * is it possible to pull the contents of `.env` into `docker-compose.yml`?
> 
>     * is it possible to pull the commands in `startup.sh` into `Dockerfile`?
> 
> 
> i find the solution referenced in [my comment](https://github.com/RWTH-i5-IDSG/steve/issues/142#issuecomment-503620485) a little bit better since it is more compact. and therefore i would like to ask @lingster the following: what did you see in that solution as problematic and wanted to improve?

goekay:
i agree that maven profiles are better than using `sed`. i don't have a problem with that. i also see the benefit of using `dockerize`. i am still questioning the necessity of `startup.sh` and whether its contents can be put in the `Dockerfile` (maybe chaining the commands with `RUN` or `CMD`)?

the better way to build the project would be:
```
mvn clean package -Pdocker -DskipTests -Djdk.tls.client.protocols="TLSv1,TLSv1.1,TLSv1.2"
```

and then run it with:
```
java -jar target/steve.jar
```

goekay:
@lingster i am taking a final look before merging this PR. thanks once again for your contribution and being tolerant with our requests! final remarks and nitpicks:
- in `main.properties` it would be better if the last two lines were changed to these:
```
db.sql.logging = false
profile = docker
```

- in `log4j2.xml`  it would be better when the root logger were in `INFO` level. please change the level in `<AsyncRoot level="DEBUG">`. otherwise logging would be too verbose.

- when you are finished with your changes please squash the commits and give me a heads up for me to merge the PR.

goekay:
i guess i can make the aforementioned changes.

PR #480, opened 2021-01-19
Cache Maven dependencies in Dockerfile

ayakovlenko:
Add an extra step to the Dockerfile to resolve Maven dependencies.


csamsel:
what do you want to achieve with that change?

PR #528, opened 2021-03-07
clean up javax.* dependencies

goekay:
the switch from `javax.*` to `jakarta.*` is happening and causing issues in steve. this PR cleans stuff up and migrates, where possible, to jakarta.


goekay:
the rest of the `javax.*` usages come from transient dependencies (e.g. jetty, jooq). we should revisit switching to jetty 10.x or 11.x in near future. however, the changes are of breaking kind.

PR #541, opened 2021-03-21
add codeql configuration back

goekay:
codeql cannot build the project because of missing database. therefore, until finding a solution i temporarily disabled/removed this feature in order to fix the master.

FYI @csamsel


csamsel:
Ah ok. Bit surprised that codeql needs to build the project for static analysis.

PR #543, opened 2021-03-26
replace go-offline command

goekay:
reason: the command cannot find dependencies outside the maven central
(our case: jitpack). it seems to be a common understanding that this
command is broken.
- same problem different context:
  https://www.mail-archive.com/issues@maven.apache.org/msg161649.html
- alternative plugin that presents go-offline's shortcomings:
  https://github.com/qaware/go-offline-maven-plugin

this PR will unblock https://github.com/RWTH-i5-IDSG/steve/pull/493

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: java
 
 
 # install dependencies
-install: mvn dependency:go-offline -s .travis-maven-settings.xml -B -V
+install: mvn dependency:resolve dependency:resolve-plugins -s .travis-maven-settings.xml -B -V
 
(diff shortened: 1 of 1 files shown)


goekay:
moreover, we are not strictly going offline anways. see: https://github.com/RWTH-i5-IDSG/steve/blob/0f0a3bdc29e995dc668e50a84e4ace4cd39505c4/.travis.yml#L22

the command there does not include `-o` to signal maven to work offline. therefore, `dependency:go-offline` was an unnecessary hurdle.

PR #663, opened 2021-09-19
Migrate to GitHub actions

goekay:
some time ago travis switched their domains which needed to renew the 3rd party integration (which we did not). therefore automated builds were not working. i hesitate to accept jooq and flyway version update PRs without having a complete build matrix across different versions.

github's built-in actions enables streamlining the builds without 3rd party integrations. i prepared the yml file with a feature parity (wrt. build matrix) to travis.yml.

when we can accept this, i will delete the leftovers of travis from the repo.

@csamsel what do you think?


csamsel:
Hi Sevket, i agree having everything in one place is nice. Also travis had some stability and security issues in the past. So i'd say lets switch to github actions. 

But i dont agree that the yaml in this PR is feature-wise on par. It does not check if the application started and is running and the matrix is smaller. Travis does Ubuntu 16.04, 18.04, 20.04, i think we can skip 16.04, but we should cover 20.04 as it's the current LTS version people will start with.

goekay:
hey @csamsel, i updated the action plans. please take another look (also at the breakdown of each step of the [runs](https://github.com/RWTH-i5-IDSG/steve/actions)).

csamsel:
yeah fine for me. Thanks

PR #819, opened 2022-05-04
Add recommended, additional configuration keys for OCMF

mhei:
The OCMF specification by SAFE e.V. recommends in the Best Practices part
three new configuration keys for OCPP, see
https://github.com/SAFE-eV/OCMF-Open-Charge-Metering-Format/blob/master/OCMF-de.md

Add these keys to the list, so that users do not need to enter it manually.

It should not harm, since OCPP clients which do not know these keys should
handle it gracefully.

Signed-off-by: Michael Heimpold <mhei@heimpold.de>


goekay:
hey @mhei thanks for your contribution. i see that you overwrote your commit, but my suggestions are still not part of that. in fact, the new commit looks the same. what changed?

mhei:
You're right, I just rebased my commit because I saw that you added something about PR and CI interaction; to be honest, I did not see your suggestions nor I received a notification by email... I'd be happy if you could repeat your suggestions again.

goekay:
oh ok, here is my comment:

> i would suggest to leave a hint to the UI user that this key is specific to OCMF. maybe modifying the `valueType` field and adding "specific to OCMF" at the end, e.g. `string, specific to OCMF`. this is only used for web page rendering and should not affect the logic.

PR #821, opened 2022-05-12
Adapt two files to project style (no functional change)

mhei:
This filters the following files with dos2unix and removes executable bit:

src/main/java/de/rwth/idsg/steve/utils/ConnectorStatusCountFilter.java
src/main/java/de/rwth/idsg/steve/utils/ConnectorStatusFilter.java

Signed-off-by: Michael Heimpold <mhei@heimpold.de>


goekay:
hey @mhei what are the effective changes in this PR? if i look at the red/green diffs, everything seems to have changed, even though content-wise both versions (before and after) seem to be the same.

mhei:
You are right, you cannot see a real difference because there is none: when I cloned the repo on my Linux PC, I see that all files have "usual" file permissions 0644 and have Linux linefeeds (LF only). Only the mentioned two files are checked in with CRLF (aka Windows/DOS style) and with permissions 0755 (here the executable bit is set). While this is not a big issue and also should not affect functionality at all, I thought it would be better have a common style throughout the complete repository.

goekay:
alrighty, thanks for the clean-up.

PR #822, opened 2022-05-12
Generate endpoint info on About page based on provided HTTP host header

mhei:
Showing all the interface addresses we're listening on is not helpful
on the About page. However, we serve the About page, we already know
that we are reaching the server using the given hostname in the
corresponding HTTP header, so let's use it to construct the URLs.

This takes also into account that the request/access already passed
proxy server(s), uses DNS CNAMEs etc.

**Warning:** Since I'm not a Java developer, I marked this PR as draft. I'm also unhappy
about some code duplications I introduce here. The primary goal is to show your
my understanding on how things should work in my eyes and to start a discussing about it.


csamsel:
It is typical that charging stations are on a different network then browser (charging station on a M2M/VPN connection, browser on internet), thats why we show all interfaces

mhei:
I fully understand your points. However, in my setup, SteVe is listening on localhost only and proxied by an Apache server - no M2M/VPN tunnel involved. In such a setup, only the localhost interface is listed in the About page which is more confusing than useful.
Maybe an idea would be to just add the HTTP host header based addresses to the existing list?

goekay:
@mhei if okay, i want to take over and finalize this. there are some things bugging me for years and i want to get rid of them :)

goekay:
i decided after this many years that i do not care about multiple network configurations. it creates complexity in `JettyServer.java` (trying to infer all hosts, which is very hit and miss IMO). if someone has some advanced setup, that person should also be able to extract the necessary info without relying on what we render on the about page.

PR #847, opened 2022-06-23
Add new notification features

juherr:
Add new notification feature about ChargePoint, User and OccpTag create/update/delete.

The pull request is not yet finished because I'd like to add some other notifications (see TODO).

I have some doubt about the naming too.

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/NotificationFeature.java
+++ b/src/main/java/de/rwth/idsg/steve/NotificationFeature.java
@@ -35,8 +35,23 @@ public enum NotificationFeature {
     OcppStationWebSocketConnected(" a JSON charging station connects"),
     OcppStationWebSocketDisconnected(" a JSON charging station disconnects"),
     OcppTransactionStarted(" a charging station starts a transaction"),
-    OcppTransactionEnded(" a charging station ends a transaction");
-
+    OcppTransactionEnded(" a charging station ends a transaction"),
+    //
+    // Charge Point related
+    //
+    ChargePointCreated(" a charging station has been created"),
+    ChargePointDeleted(" a charging station has been deleted"),
+    ChargePointUpdated(" a charging station has been updated"),
+    UnidentifiedChargePointProceed(" an unidentified charging station has been proceed"),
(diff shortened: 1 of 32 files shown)


goekay:
thanks for this! i know it's a draft, but here is the one thing that occurred to me after glancing over it:

there are too many dumb data holder classes who share similar fields and their only reason for existence is for the listener to catch them by the class name. can we unify them? something like:

```
@Data
public class ChargePointOperation {
  private final int chargeBoxPk;
  private final String chargeBoxId;
  private final NotificationFeature type;
}
```

and then the listener can be:
```
@EventListener(condition = "#notification.type == NotificationFeature.ChargePointDeleted")
public void chargePointDeleted(ChargePointOperation notification) {
   // do some cool logic
}
```
or, if we want to reduce spring magic::
```
@EventListener
public void chargePointDeleted(ChargePointOperation notification) {
    if (notification.type == NotificationFeature.ChargePointDeleted) {
        // do some cool logic
    }
}
```

juherr:
You're right about the type of notifications.

All notifications on the same item share the same description because I don't know yet what is helpful to set.
I don't like adding `NotificationFeature type` in the class because it will allow having a tag feature on a charging station notification.

But I will try the following:
* a generic notification with `NotificationFeature type` + `int pk` (the type determines the reference of the pk)
* custom notifications when needed

It's a draft because not ready for merge but open to remarks :) 
Do you have some others?

Fyi, my next PR will be to add a `WebhookNotificationService` for a Slack (or other) integration

juherr:
About naming, what is the difference between `OcppStation` and `ChargePoint`?

juherr:
Ping @goekay 
Please, advice

juherr:
update: not dead but less free time on my side ;)

anirudh-ramesh:
Hi @juherr, I've experimented with adding webhooks that were called from NotificationService. It's a hack since the WebhookService was called from methods of MailService instead of NotificationService. I also got a chance to add some columns to the MySQL database for storing a single webhook link along with enable/disable flags. I need to clean it up and refactor it in a manner suitable for adding to SteVe's master. But, similarly, not dead but less free time :/

PR #953, opened 2022-10-30
Use service for writing operations

juherr:
With the new API controllers, the business logic of writing operations should be shared and the service layer is a good place for that.

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/repository/OcppTagWriteRepository.java
+++ b/src/main/java/de/rwth/idsg/steve/repository/OcppTagWriteRepository.java
@@ -18,29 +18,16 @@
  */
 package de.rwth.idsg.steve.repository;
 
-import de.rwth.idsg.steve.repository.dto.OcppTag;
-import de.rwth.idsg.steve.web.dto.OcppTagQueryForm;
-import jooq.steve.db.tables.records.OcppTagActivityRecord;
-import org.jooq.Result;
-
+import de.rwth.idsg.steve.web.dto.OcppTagForm;
 import java.util.List;
 
 /**
  * @author Sevket Goekay <sevketgokay@gmail.com>
  * @since 19.08.2014
  */
-public interface OcppTagRepository {
-    List<OcppTag.Overview> getOverview(OcppTagQueryForm form);
-
-    Result<OcppTagActivityRecord> getRecords();
-    Result<OcppTagActivityRecord> getRecords(List<String> idTagList);
-
-    OcppTagActivityRecord getRecord(String idTag);
-    OcppTagActivityRecord getRecord(int ocppTagPk);
-
-    List<String> getIdTags();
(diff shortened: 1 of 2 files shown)


juherr:
Ping @goekay 
What do you think about this design?

goekay:
hey @juherr, sorry, life kept me busy last couple of weeks.

i like the motivation! i think i know where you are coming from (deduplication of removeUnknown(..) calls?) but i would like to minimize the code ceremony and verbosity. here, you also introduce an additional layer for update and delete operations that just forwards these calls. 

i yet have to collect my thoughts about this approach in order to give a proper feedback. will get back to you in a couple of days.

juherr:
In fact, I had some pains in my other PR because I had some difficulties to understand the architecture model and find where is the business logic: some are in service and some other in repositories usages.

The duplication is a problem here because you may easily not have the same behavior between the api and the ui.

I see 2 possible approachs to fix it:
* Use services as domain layer between repositories and controllers (something between n-tiers, ddd and hexagonal architecture) but many repository calls are just read without value add and using a service is maybe too much
* Separate read and write operations (cqrs architecture) that allows to reduce the effort and keep the repository usages in controller.

The current PR is a try for the 2nd option. My main objective is to understand what is your feeling about architecture and be able to propose some changes in this direction.

goekay:
> In fact, I had some pains in my other PR because I had some difficulties to understand the architecture model and find where is the business logic: some are in service and some other in repositories usages.

agree. background and context: 

during the initial development years of steve, my pragmatic approach was the following. many read/write operations from the controller layer can be forwarded to the repository layer since there is not much business logic going on in-between. and, since i heavily use the capabilities of sql, my repository methods are not just "read by id, and give the whole entity" (the usage of jooq enabled fluent chaining of mapping logic as well). if this were the case, it would have necessitated a service layer, where i would have mapped request and response classes and aggregating info from many repository calls and so on. however, this was not the case and i found an artificial middle service layer... artificial? (some repository methods are quite complex though. you might argue that they contain business logic, and i might/would agree ;)) 

over time, when new business requirements emerged, i introduced service classes where necessary since these requirements did not fit controller-repository flows anymore. and this was done selectively and not in a general manner. the ugliness of this approach and the confusion it created were bearable since i was the only developer :) there are two noteworthy examples of this problem: OcppTagService/OcppTagRepository and ChargePointHelperService/ChargePointRepository. you address the first one.

---

after some inspection, i came to the conclusion that a clear "controller -> service -> repository" pattern would be better for the ocpp tag flows even though it adds just forwarding in some flows. the consistency would outweigh. therefore, i would be for your first option.

if you want to work on it, i would be glad. otherwise, i would do it. and you can change the target to master instead of this feature branch. the feature branch might wait a little longer. i would merge your contribution ASAP.

juherr:
Changes were already in the master branch and read/write separation is not important.

Issue #963, opened 2022-11-11
Charging profile assignment is not properly removed


lategoodbye:
#### Checklist

- [x] I checked other issues already, but found no answer/solution
- [x] I checked the documentation and wiki, but found no answer/solution
- [x] I am running the latest version and the issue still occurs
- [x] I am sure that this issue is about SteVe (and not about the charging station software or something unrelated to SteVe)

#### Specifications
```
SteVe Version     : 3.5.0
Operating system  : Debian 11
JDK               : OpenJDK 17.0.4
Database          : Maria DB 10.5
```

#### Expected Behavior

In case SteVe gets informed (ClearChargingProfile.Conf) that a specific charging profile is not active anymore, the charging profile assignment to a specific charge point should be dropped.

#### Actual Behavior

In some corner cases the charge profile assignment persists and prevents modification on that charging profile.

de.rwth.idsg.steve.SteveException: Cannot modify this charging profile, since the following stations are still using it: ...

#### Steps to Reproduce the Problem

Corner case charging profile is cleared locally on the charging station:
  1. Install charging profile on Charge point via SetChargingProfile
  2. Remove charging profile locally
  3. Clear the charging profile on Charge point via ClearChargingProfile with ID
  4. Charge point returns "Unknown" but Steve keeps the assignment

goekay:
> In case SteVe gets a notification that a specific charging profile is not active anymore

when is this happening? and why?

lategoodbye:
Notification is a bad wording, i mean message. As discussed in the pull request in OCPP 1.6 there are only two types of response status (Accepted or Unknown).

goekay:
@lategoodbye can we close this issue since the PR is merged?

PR #968, opened 2022-11-17
Remove charging profile regardless of reply status #963

lategoodbye:
In case the charge point doesn't known the charging profile that should be cleared, it might be already done. So remove the profile regardless of the reply status from the repository to be in sync.


goekay:
thanks @lategoodbye!

my knowledge about this part of ocpp was a little bit rusty, so i revisited the spec. the station's response can only be "Accepted" or "Unknown" _(No Charging Profile(s) were found matching the request)._ a station cannot reject a ClearChargingProfile request. this is really interesting. therefore, your justification for deletion resonates with me. 

however, i would make a suggestion: it would be better to have log line that we are deleting the profile regardless of the answer. otherwise, it might confuse users. something like:

```
// https://github.com/steve-community/steve/pull/968
if (!"Accepted".equalsIgnoreCase(statusValue)) {
    log.info("Deleting the entry from DB even though the response was {}", statusValue);
}
```

lategoodbye:
Okay, i will try to test the suggestion and push it

lategoodbye:
Commit reworked

goekay:
thank you!

PR #984, opened 2022-12-03
[OpenAPI] Better describe how the From-To Filter works

kohlerdominik:


goekay:
thanks for the update!

QueryForm is used for transactions but also for reservations. Therefore, the wording "transmissions" would be misleading.

PR #988, opened 2022-12-08
Fix the objectMapper bug and refactor properties for k8s

ByungJun25:
Hi steve teams!
Thank you for creating good open-source.

I would like to fix the bug and also refactor properties for kubernetes.

You can reproduce the bug like below.

1. Run the application with `webapi.value` for exposing api endpoints.
2. Send the get request with invalid key value in header.
3. You will get the 500 error instead of 401 error.

The reason why it happened is that the jackson library doesn't support joda module as default. So I registered joda module as manually when the Object mapper was created.

And also I refactored properties of kubernetes, because k8s environment should be get the value dynamically by enviroment value.

Please review and give me feedback. 
Thanks.

### DIFF
--- a/k8s/yaml/Deployment.yaml
+++ b/k8s/yaml/Deployment.yaml
@@ -32,3 +32,7 @@ spec:
           value: ""
         - name: ADMIN_PASSWORD
           value: ""
+        - name: WEBAPI_KEY
+          value: "STEVE-API-KEY"
+        - name: WEBAPI_VALUE
+          value: ""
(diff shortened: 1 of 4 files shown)


goekay:
hey @ByungJun25 good catch on both parts, and thanks for your contribution!

background on ObjectMapper (for people who like details): the issue was introduced with ba788603412c79bf283fa216331d8bcd1bfc0ed4 where i also modified the data type for ApiErrorResponse timestamp from String to DateTime. ApiErrorResponses for normal API calls can still be deserialized without error, because the ObjectMapper instance for "normal calls" comes from spring's MappingJackson2HttpMessageConverter. spring detects JodaModule automatically via classpath scanning and registers it.

however, in ApiKeyAuthenticationEntryPoint we create a local ObjectMapper instance (which is disconnected from spring's context) and use it instead.

i will try to find ways to get both aligned.

goekay:
i made an improvement pass with the motivation outlined in the message above. i hope you are okay with that. thanks again for pointing out this issue.

ByungJun25:
@goekay Thank you for explaining details and solving problem :+1:

PR #992, opened 2022-12-10
Refactor OcppTagService

goekay:
simplify some internal code in OcppTagService, and use OcppTagService for all things related to ocpp tags. do not expose OcppTagRepository to controller layer. it is only used internally by OcppTagService.

context: https://github.com/steve-community/steve/pull/953#issuecomment-1321275033


goekay:
hey @juherr what do you think about 6f5c28487fd578b294913f099be196eb60d7747f ?

PR #998, opened 2022-12-11
Add maven wrapper

juherr:


juherr:
Ping @goekay 
WDYT?

The documentation of the maven wrapper is here: https://maven.apache.org/wrapper/
Using it is a good way to enforce a specific version and help developers to use the same version without pain.

goekay:
agreed. thanks @juherr !

PR #1002, opened 2022-12-11
Rebased remove_db_build branch

juherr:


juherr:
Will be fixed with https://github.com/steve-community/steve/issues/1733

PR #1013, opened 2022-12-18
Display all events transaction if asked

### LINKED ISSUE #1004: [Question] Why does transaction details show only the first meter value?
#### Checklist

- [X] I checked other issues already and found no answer
- [X] I checked the documentation and found no answer
- [ ] I am running the latest version and the feature i am requesting is not implemented there

Hi, Steve team!

first, I don't know about OCPP spec enough, so it is possible to ask with wrong question..
This question is similar with this [issue](https://github.com/steve-community/steve/issues/97). 
I need to show real time power of connector.

I just noticed now, the transaction detail page shows only the first meter value.
https://github.com/steve-community/steve/blob/55c7212f6f8f812c215c9e05d580fbbae018c8ea/src/main/java/de/rwth/idsg/steve/repository/TransactionRepository.java#L42

And according to comments of codes, it said that the meter value is not changeable, so your team showed only one data.
https://github.com/steve-community/steve/blob/55c7212f6f8f812c215c9e05d580fbbae018c8ea/src/main/java/de/rwth/idsg/steve/repository/impl/TransactionRepositoryImpl.java#L179-L185

But it used last meter value for stopping transaction.
https://github.com/steve-community/steve/blob/55c7212f6f8f812c215c9e05d580fbbae018c8ea/src/main/java/de/rwth/idsg/steve/service/TransactionStopService.java#L59-L61

Well, I wonder why it shows only first meter value on transaction detail page and how I can get the real-time power correctly.
Thank you.

juherr:
Fix #1004

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/repository/TransactionRepository.java
+++ b/src/main/java/de/rwth/idsg/steve/repository/TransactionRepository.java
@@ -38,6 +38,10 @@ public interface TransactionRepository {
 
     TransactionDetails getDetails(int transactionPk, boolean firstArrivingMeterValueIfMultiple);
 
+    /**
+     * @deprecated Use {@link #getDetails(int, boolean)} instead
+     */
+    @Deprecated
     default TransactionDetails getDetails(int transactionPk) {
         return getDetails(transactionPk, true);
     }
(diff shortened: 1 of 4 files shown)


goekay:
hey, thanks for the PR but this is not as easy as this.

the flag is used at https://github.com/steve-community/steve/blob/master/src/main/java/de/rwth/idsg/steve/repository/impl/TransactionRepositoryImpl.java#L187-L192

if the flag is false, it will select the last one. we/i need to change the sql structure to allow selecting all events.

moreover, for the flag to be effective the JSP file needs to be changed with additional UI constructs for the user to interact with this property.

juherr:
Thank for the highlight, I didn't check the repository layer. Inwill update.

> moreover, for the flag to be effective the JSP file needs to be changed with additional UI constructs for the user to interact with this property.

The idea was to hide the option and not show it in the ui: power user can add the flag in the url.
But I can add an url in the view if needed.

goekay:
what is the state of this PR? especially, after removing firstArrivingMeterValueIfMultiple?

juherr:
Should be reworked

PR #1026, opened 2023-01-08
add specific pmd version

goekay:
reason: github action implicitly uses 6.53.0. the rulesets from maven-pmd-plugin-3.19.0
are for 6.49.0. the action was failing with "Error: Invalid rulesets"


goekay:
FYI @juherr

PR #1027, opened 2023-01-08
remove pmd job in github actions

goekay:
the behaviour of this github action seems to be flaky.

master was broken after adding this job: https://github.com/steve-community/steve/actions/runs/3836455431/jobs/6587998999

based on the github docs, i hoped/assumed that the syntax for the `sourcePath` value had to be a different one. i attempted a solution with: https://github.com/steve-community/steve/commit/f96b9e8e57a6431aea458d886ed2d33f22e38d19

this fixed master. after rebasing a feature branch, the branch was broken. attempted fix: https://github.com/steve-community/steve/pull/1026

the branch still fails: https://github.com/steve-community/steve/actions/runs/3865067797/jobs/6588324097

all of this is too much headache for little value.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -2,23 +2,6 @@ name: check quality
 on: [ push, pull_request ]
 
 jobs:
-  pmd:
-    runs-on: 'ubuntu-latest'
-    steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-java@v3
-        with:
-          distribution: 'temurin'
-          java-version: '17'
-      - name: Run PMD
-        uses: pmd/pmd-github-action@v1
-        with:
-          version: '6.49.0'
-          sourcePath: './src/main/java'
-          rulesets: 'https://raw.githubusercontent.com/apache/maven-pmd-plugin/maven-pmd-plugin-3.19.0/src/main/resources/rulesets/java/maven-pmd-plugin-default.xml'
-      - name: Fail build if there are violations
-        if: steps.pmd.outputs.violations != 0
-        run: exit 1
   checkstyle:
     runs-on: 'ubuntu-latest'
     steps:


goekay:
@juherr what do you say?

goekay:
in the same spirit, i want to get rid of license checks as well, because: https://github.com/steve-community/steve/actions/runs/3865067796/jobs/6588432634

if you inspect `ApplicationJsonTest` and `OcppWebSocketHandshakeHandler`, you can see that the license headers are there. if i run `mvn -B -V license:check --file pom.xml` locally, i don't see any failures. i have no idea why the behaviour would be different with github actions.

juherr:
That's strange. Let me some time to understand the issue.
At least, you can keep the actions for pull requests only because they help contributors and you to review changes.

goekay:
any update?

juherr:
Did you test with the same environment as CI? OS, JDK, and Maven?

To avoid the potential problem with maven, the Maven wrapper exists to enforce the version.
Could you consider adding it? See https://github.com/steve-community/steve/pull/998

juherr:
About PMD, the action (and maybe the cli) doesn't support http as ruleset path: https://github.com/pmd/pmd-github-action/blob/5539369540aad308d8e2cff0600ac6d6943dce99/lib/validator.js#L25 -> `:` is not supported but included in `https://`.
Fun fact: the maven plugin supports http

juherr:
I reproduced the license check issue locally and the fix is included in https://github.com/steve-community/steve/pull/1030
I just don't understand why only these 2 files are impacted.

goekay:
thanks.

regarding the license header: the rest of the files have `2013-2022` which does not violate the license checks. weird that only 2 files have that issue, and your change resolves that while other fly under the radar. where does this `${license.git.copyrightYears}` come from?

goekay:
if i delete all headers and regenerate them using the plugin, all files have `2013-2023`.

juherr:
`${license.git.copyrightYears` comes from the plugin which is based on git.
I didn't check the implementation but I think it check the last time the file was modified.
If you made a modification in this 2 files in 2023, it explain the issue (but not why you didn't reproduce).

On that subject, I found some discussions about the years to apply in the licence.
Some say that only the second year should be set because there is no sense to have a past year for new files.

juherr:
> if i delete all headers and regenerate them using the plugin, all files have `2013-2023`.

It confirms what I said: if you delete headers, then you modify files in 2023. So the applied license will 2023.

goekay:
i did not modify these two files in 2023

PR #1088, opened 2023-03-17
Allow Heartbeat Interval to be zero

mhei:
According to the standard, when a value of zero is returned in the BootNotification, then the client is requested to use a reasonable default interval:

Quote:
  If that interval value is zero, the Charge Point chooses a
  waiting interval on its own, in a way that avoids flooding
  the Central System with requests.

So let's allow the configuration of zero here.


goekay:
Nice, thanks @mhei!

Issue #991, opened 2022-12-10
Support multi users and rbac


juherr:
#### Checklist

- [x] I checked other issues already and found no answer
- [x] I checked the documentation and found no answer
- [x] I am running the latest version and the feature i am requesting is not implemented there

#### Describe the problem you are trying to solve

Currently, only one user is supported with all privileges. It is not a problem for a private instance but becomes dangerous when the instance is shared (especially the delete operations).

#### Describe the solution you'd like

Replace the mono static user by users stored in database.
It should be possible to enable or disable any crud operation one by one by user or role.

#### Describe alternative solutions or features you've considered 

The user storage could be delegate to a 3rd party backend like okta or any other saas solution.

#### Additional context

...

goekay:
> It is not a problem for a private instance but becomes dangerous when the instance is shared (especially the delete operations).

for a proper solution, we should have auditing of modifying operations. that way, an admin can track which user did what. i don't want to open pandora's box and cause more work within current PRs, but something for future.

juherr:
The basic REST lifecycle for manager/API accounts is now tracked separately by #2075. This issue remains the source of truth for multi-user authorization, granular permissions, RBAC, and auditing.

PR #1137, opened 2023-05-02
feat: add oidc authentication to allow sso

fritterhoff:
Hey! :)

I just stumbled over this cool project and we plan to utilize it for the charging infrastructure at our campus. Since we don't want to share the admin password to several users and we would prefer to reuse existing credentials and MFA I implemented a POC version of an OIDC authentication in steve. For sure it is not perfect but we would like to contribute our adaptions to the official project 😉

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -592,7 +592,16 @@
             <artifactId>spring-security-config</artifactId>
             <version>${spring.security.version}</version>
         </dependency>
-
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-oauth2-client</artifactId>
+            <version>${spring.security.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-oauth2-jose</artifactId>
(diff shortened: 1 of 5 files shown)


juherr:
Related to #991

fritterhoff:
Some (long) time ago I created this PR. This is getting relevant again for us. I would consider to update the conflict. Are there any plans to integrate such a feature in the near future?

goekay:
sure, lets reactivate this (there are many PRs that need to be revitalized :D) 

i assume the code changes work without issues (functionally), but i am not sure whether the `if ... else ...` approach is the best solution in terms of modularity. can you take a look and maybe rethink that aspect of the solution?

goekay:
another point: would you like to wait a bit for the project to migrate to spring boot? this migration is in the TODO list with a moderate urgency/priority. after the migration, i think you could use config params from spring security directly instead of reinventing the wheel in steve's own application properties.

fritterhoff:
Would be fine for us!

goekay:
@fritterhoff we are using spring boot now. feel free to proceed with the preparation.

PR #1140, opened 2023-05-08
Use testcontainers for jooq generation setup

### LINKED ISSUE #956: Add V1 baseline migration in order to avoid SUPER privilege need from 0.x schema version
#### Checklist

- [x] I checked other issues already and found no answer
- [x] I checked the documentation and found no answer
- [x] I am running the latest version and the feature i am requesting is not implemented there

When you build Steve, you need to have a database with SUPER privilege otherwise you are not able to run migrations (see #495 ad other related issues).
This need was removed in a later migration and V1.x doesn't need it anymore.

#### Describe the solution you'd like

Flyway has baseline migration which skips migrations files when not needed.
https://flywaydb.org/documentation/tutorials/baselineMigrations
Having a V1.0 baseline migration will skip the migrations which need more privileges.

#### Describe alternative solutions or features you've considered 

Do it by hand:
1. Run migrations on a database with enough privilege
2. Export schema
3. Import schema into the target database without super privilege

#### Additional context

When working, it will be possible to remove the FAQ entry.
https://github.com/steve-community/steve/wiki/FAQ#database-migration-fails-on-mariadb-due-to-missing-super-privilege

juherr:
To enable it, just add run:
`mvn -Pdev,testcontainers package`

Optionally, all V0 migrations are merged into the same baseline and can be used to avoid the need to grant privileges to the db user (Fix #956).

To enable it, just add run:
`mvn -Pdev,testcontainers -Ddb.baseline=V1 package`

It should help new users to bootstrap the project more easily.


juherr:
Ping @goekay 
WDYT?

juherr:
Try the very new plugin instead https://testcontainers.com/guides/working-with-jooq-flyway-using-testcontainers/

juherr:
Partially replaced by #1455

PR #1162, opened 2023-05-30
Update transactions.jsp

fnkbsi:
The STOP button in Transaction.jsp indicates that the loading process can be stopped with the button. The underlying functionality creates a stop event entry in the SQL table, but does not remotely stop the transaction (remoteStopTransaction). Changed the stop button title to better understand this behavior.


goekay:
very good point, thanks @fnkbsi !

goekay:
history and context: this button was added to mark some "ghost" transactions as stopped.

PR #1164, opened 2023-05-31
TransactionDetails data reduction with lag and lead and export to csv

### LINKED ISSUE #1004: [Question] Why does transaction details show only the first meter value?

fnkbsi:
Adding an export function of the transaction details to provide easy access to the data of the charging process.
The use of data reduction is optional. The reduction algorithm is changed from first/last ever appearance of equal values to first/last appearance of equal values behind each other.
May fix #1004


fnkbsi:
The flag firstArrivingMeterValueIfMultiple is removed at the master, so this PR seems obsolete!

PR #1165, opened 2023-06-02
Change from single user (config) to database based multi user system.

### LINKED ISSUE #991: Support multi users and rbac

fnkbsi:
Implements multi user access. Uses the steve database to store the "webusers" and their roles. Privileges on websites are set in the  source code in SecurityConfiguration.java. Sites for administration of the "webuser" included.
May fixes #991


pom.xml
```diff
@@ -772,5 +772,17 @@
             <artifactId>encoder</artifactId>
             <version>1.2.3</version>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
```
juherr, line 776: It looks the 2 new dependencies are not used. They should be removed IMO.

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
```diff
@@ -38,13 +39,13 @@
 import org.springframework.security.config.http.SessionCreationPolicy;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
-import org.springframework.security.core.userdetails.User;
-import org.springframework.security.core.userdetails.UserDetails;
-import org.springframework.security.core.userdetails.UserDetailsService;
-import org.springframework.security.crypto.factory.PasswordEncoderFactories;
-import org.springframework.security.crypto.password.DelegatingPasswordEncoder;
+//import org.springframework.security.core.userdetails.User;
```
juherr, line 42: Commented code doesn't the review. Could you remove the code instead?

juherr (commented):
The feature looks nice but I didn't test yet.

Only @goekay will be able to merge it.

Please consider my remarks because they will help him to review and give feedbacks.

pom.xml
fnkbsi: Both dependencies used in SecurityConfiguration.java 
by "import org.springframework.boot.jdbc.DataSourceBuilder;" (line 55)

goekay: steve is not using spring boot though. it is spring app, not a spring boot app.

juherr: @fnkbsi steve uses jooq for the data access. You should be able to replace spring-jdbc by that. Then no need to datasourcebuilder.

juherr: @goekay any feedback about the feature itself?

src/main/java/de/rwth/idsg/steve/config/BeanConfiguration.java
```diff
@@ -82,13 +84,14 @@
 @ComponentScan("de.rwth.idsg.steve")
 public class BeanConfiguration implements WebMvcConfigurer {
 
-    private HikariDataSource dataSource;
+    @Autowired private HikariDataSource dataSource;
```
juherr, line 87: You should remove the attribut and replace it by the bean factory method everywhere it is used (spring proxies are not supposed to instanciate more than once)

src/main/java/de/rwth/idsg/steve/config/BeanConfiguration.java
```diff
@@ -127,7 +130,9 @@ private void initDataSource() {
      */
     @Bean
     public DSLContext dslContext() {
-        initDataSource();
```
juherr, line 130:
```java
HikariDataSource dataSource = initDataSource(); // better if named dataSource()
```

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
```diff
@@ -66,29 +65,27 @@
...
+    
     /**
      * Password encoding changed with spring-security 5.0.0. We either have to use a prefix before the password to
      * indicate which actual encoder {@link DelegatingPasswordEncoder} should use [1, 2] or specify the encoder as we do.
      *
      * [1] https://spring.io/blog/2017/11/01/spring-security-5-0-0-rc1-released#password-storage-format
      * [2] {@link PasswordEncoderFactories#createDelegatingPasswordEncoder()}
+    */
+
+    /**
```
juherr, line 80: Useless change

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
```diff
@@ -100,23 +97,90 @@ public WebSecurityCustomizer webSecurityCustomizer() {
     @Bean
     public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
         final String prefix = CONFIG.getSpringManagerMapping();
-
         return http
-            .authorizeHttpRequests(
-                req -> req.antMatchers(prefix + "/**").hasRole("ADMIN")
-            )
+            .authorizeRequests(
```
juherr, line 101: What are the expected rights?

On my side, I'd love that users can read but not create/update/delete.
Admin can do everything.

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
```diff
@@ -100,23 +97,90 @@ public WebSecurityCustomizer webSecurityCustomizer() {
...
+                 req -> req
+                    .antMatchers(prefix + "/home").hasAnyRole("USER", "ADMIN")
+                    .antMatchers(prefix + "/users/" + "**").hasAnyRole("USER", "ADMIN")
+                    .antMatchers(prefix + "/ocppTags/" + "**").hasAnyRole("USER", "ADMIN")
+                    .antMatchers(prefix + "/signout/" + "**").hasAnyRole("USER", "ADMIN")
+                    .antMatchers(prefix + "/noAccess/" + "**").hasAnyRole("USER", "ADMIN")
+                    .antMatchers(prefix + "/**").hasRole("ADMIN")
```
juherr, line 108: Don't forget to manage the api route

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
```diff
@@ -100,23 +97,107 @@ public WebSecurityCustomizer webSecurityCustomizer() {
...
+    }
+
+
+    /**
+     *
+     * @return
+     */
+    @Bean
+    public UserDetailsManager authenticateUsers() {
+      JdbcUserDetailsManager users = new JdbcUserDetailsManager(dataSource);
```
juherr, line 170: To be discussed but I think a home-made `JooqUserDetailsManager` is not a difficult task and will avoid an extra dependency.

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
fnkbsi: The api route is managed by the api key security filter chain. 
A similar approach for the api key or a combination of api key and user/password is possible, but needs to be discussed.

juherr: Right. I forgot it 😅

juherr:
@goekay Please, review. The change will improve the product a lot.

fluppie:
nice, hope this will be merged soon :-)!

goekay:
this PR is next on my TODO list. starting to review it.

goekay:
@fnkbsi could you pls undo useless/whitespace/stylistic changes which would make it easier for me to see the _real_ changes?

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
goekay: agreed. it could then replace `WebUserRepositoryImpl` and some of the other classes.

i think a direct Jooq impl of UserDetailsManager, UserDetailsService that use the classes UserDetails (and User) would nicely align this impl with the concepts of Spring... and therefore even make the home-made classes such as WebUser redundant.

src/main/resources/db/migration/V1_0_6__update.sql
```diff
@@ -0,0 +1,29 @@
+-- adding tables for multi users access
+
+CREATE TABLE `webusers` (
```
goekay, line 3: suggestion: `web_users` since we use underscore as delimiter.

src/main/resources/db/migration/V1_0_6__update.sql
```diff
@@ -0,0 +1,29 @@
+-- adding tables for multi users access
+
+CREATE TABLE `webusers` (
+  `username` varchar(50) COLLATE utf8mb3_unicode_ci NOT NULL,
+  `password` varchar(255) COLLATE utf8mb3_unicode_ci NOT NULL,
+  `enabled` tinyint(1) NOT NULL,
+  PRIMARY KEY (`username`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_unicode_ci;
+
+CREATE TABLE `webauthorities` (
```
goekay, line 10: suggestion: `web_user_authorities`

src/main/java/de/rwth/idsg/steve/repository/impl/WebUserRepositoryImpl.java
```diff
@@ -0,0 +1,279 @@
...
+ * @author fnkbsi
+ * @since 01.04.2022
+ */
+@Slf4j
+@Repository
+public class WebUserRepositoryImpl implements WebUserRepository {
+
+    @Autowired private DSLContext ctx;
+
+    private PasswordEncoder encoder = new BCryptPasswordEncoder();
```
goekay, line 58: you should use [standard Bean PasswordEncoder](https://github.com/steve-community/steve/blob/master/src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java#L77)

src/main/resources/db/migration/V1_0_6__update.sql
goekay: opinion: i prefer to use singular names when naming tables... because obviously every table would have an `s` (or whatever) at the end to denote plurality since every table would have multiple entries. if every table name is plural, this notion loses its value/meaning and you could then stick to singular.

i know this is an opinionated convention and others might have different ones, but since the db tables of steve already use the singular names... i would suggest to stay with that just for the sake of consistency, even if you dont agree with my approach.

src/main/resources/db/migration/V1_0_6__update.sql
goekay: https://github.com/steve-community/steve/pull/1165/files#r1712671917

src/main/resources/db/migration/V1_0_6__update.sql
```diff
@@ -0,0 +1,29 @@
...
+  CONSTRAINT `authorities_ibfk_1` FOREIGN KEY (`username`) REFERENCES `webusers` (`username`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_unicode_ci;
+
+CREATE UNIQUE INDEX ix_auth_username
+  on webauthorities (username,authority);
+
+-- Insert a user = admin with the password = pass. Change password after installing!
+INSERT INTO webusers (username, password, enabled)
+  values ('admin',
+    '$2a$10$.Rxx4JnuX8OGJTIOCXn76euuB3dIGHHrkX9tswYt9ECKjAGyms30W',
```
goekay, line 23: current standard password is 1234. i would suggest to pass this value through the encoder and save it.

even better: we should probably think about a migration strategy of existing installations, such as...

we can read the `CONFIG.getAuth().getUserName()` and `CONFIG.getAuth().getEncodedPassword()` and insert this user into table when during boot, if the table does not have this entry already.

src/main/resources/db/migration/V1_0_6__update.sql
goekay: maybe two tables are not needed? controversial suggestion: i know normalization is good etc. but this is really a small use case. denormalize and inject authorities into web_user as a column?

mysql supports [json](https://dev.mysql.com/doc/refman/8.4/en/json.html). this column could be a json array of string values.

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
goekay: [here is my quick attempt](https://github.com/steve-community/steve/commit/df2f9aa1bc21969426b3d6aba554214d0e43ba0d)

goekay: [addendum](https://github.com/steve-community/steve/commit/3491594b7b4898cda5d41c883cb76a8fa688a77a) to fine-tune details. it is in working condition now.

src/main/resources/db/migration/V1_0_6__update.sql
goekay: postConstruct impl in SecurityConfiguration of https://github.com/steve-community/steve/pull/1539/commits/3491594b7b4898cda5d41c883cb76a8fa688a77a does exactly this.

PR #1202, opened 2023-07-29
Use SteveConfiguration as Spring component

juherr:
Goal: Avoid global variable and inject instance everywhere needed.

Extra:
* Extract UnknownChargePointService in order to remove the dependency cycle.
* Prefer constructor injection instead of field injection when a class was modified

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/Application.java
+++ b/src/main/java/de/rwth/idsg/steve/Application.java
@@ -37,7 +37,7 @@ public class Application implements ApplicationStarter, AutoCloseable {
         // For Hibernate validator
         System.setProperty("org.jboss.logging.provider", "slf4j");
 
-        SteveConfiguration sc = SteveConfiguration.CONFIG;
+        SteveConfiguration sc = new SteveConfiguration();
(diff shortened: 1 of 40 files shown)


juherr:
Ping @goekay wdyt?

juherr:
Ping @goekay

goekay:
this is hard for me to review as there are too many changes in this PR, but also in one commit. can you split it to multiple PRs, or at least multiple commits?

regarding the constructor injection: i would welcome it in conjunction with Lombok to reduce constructor ceremony using `@RequiredArgsConstructor`. moreover, i see you did not make the changes in all classes. may i ask the reason?

thanks for the effort!

juherr:
> moreover, i see you did not make the changes in all classes. may i ask the reason?

Because it was not the purpose of the pull request.

As asked: #1226

juherr:
> this is hard for me to review as there are too many changes in this PR

Even if I rebase onto #1226, I'm not sure the change will be really shorter because the steve configuration global variable is used deep in the object graph.

juherr:
Closing this PR due to inactivity (no response in 11 months). Feel free to reopen if needed.
See: https://github.com/juherr/evolve/pull/18

Issue #1212, opened 2023-08-17
DB migrations fail in Docker with MariaDB 10.4.31


sirhcel:
#### Checklist

- [x] I checked other issues already, but found no answer/solution
- [x] I checked the documentation and wiki, but found no answer/solution
- [x] I am running the latest version and the issue still occurs
- [x] I am sure that this issue is about SteVe (and not about the charging station software or something unrelated to SteVe)

#### Specifications
```
SteVe Version     : release 3.6 and the current head at 0a2e75b3
Operating system  : Docker
JDK               : ...
Database          : MariaDB 10.4 as selected by Dockerfile
```

#### Expected Behavior

Starting up application and database containers succeeds.

#### Actual Behavior

Starting up the application container fails at running database migrations.

#### Steps to Reproduce the Problem

1. Stop and remove eventually running instances and volumes
    ```
    $ docker stop steve_app_1 steve_db_1 
    $ docker rm steve_app_1 steve_db_1
    $ docker volume rm steve_db-data
    $ docker image rm mariadb:10.4 steve_app
    ```

2. Clone repository
    ```
    $ git clone https://github.com/steve-community/steve.git
    $ cd steve
    steve$ docker-compose up
    ```

3. Running database fails and there are multiple types of issues (so far)

    1. Presence of foreign keys in an early stage of the migration
        ```
        app_1  | [INFO] Flyway Community Edition 7.15.0 by Redgate
        app_1  | [INFO] Database: jdbc:mysql://mariadb:3306/stevedb (MySQL 5.5)
        app_1  | [INFO] Successfully validated 35 migrations (execution time 00:00.013s)
        app_1  | [INFO] Creating Schema History table `stevedb`.`schema_version` ...
        app_1  | [INFO] Current version of schema `stevedb`: << Empty Schema >>
        app_1  | [WARNING] outOfOrder mode is active. Migration of schema `stevedb` may not be reproducible.
        app_1  | [INFO] Migrating schema `stevedb` to version "0.6.6 - inital"
        app_1  | [WARNING] DB: Event scheduler is switched off, use SET GLOBAL event_scheduler=ON to enable it. (SQL State: HY000 - Error Code: 1105)
        app_1  | [INFO] Migrating schema `stevedb` to version "0.6.7 - update"
        app_1  | [WARNING] DB: Event scheduler is switched off, use SET GLOBAL event_scheduler=ON to enable it. (SQL State: HY000 - Error Code: 1105)
[...]

PR #1213, opened 2023-08-17
Pin version of MariaDB for DB migrations

sirhcel:
When running database migrations with the latest 10.4.31, they fail as shown in issue #1212. Pinning the point release of MariaDB to the latest known-good version maneuvers around this issue.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ jobs:
       matrix:
         os: [ ubuntu-20.04, ubuntu-22.04 ]
         java: [ '11', '17' ]
-        db: [ 'mysql:5.7', 'mysql:8.0', 'mariadb:10.3', 'mariadb:10.4', 'mariadb:10.5', 'mariadb:10.6' ]
+        db: [ 'mysql:5.7', 'mysql:8.0', 'mariadb:10.3', 'mariadb:10.4.30', 'mariadb:10.5.21', 'mariadb:10.6.14' ]
     runs-on: ${{ matrix.os }}
     services:
       mysql:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,12 @@ volumes:
 services:
 
   db:
-    image: mariadb:10.4
+    # Pinning MariaDB to point release 10.4.30 works around the issues with the
+    # database migrations seen with 10.4.31 in issue #1212.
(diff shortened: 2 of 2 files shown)


juherr:
This change doesn't fix the issue, it is just hiding it :/

sirhcel:
Yes, you are absolutely right. Let me rephrase the title. This PR aims at shipping a configuration that just works and might save others the time for troubleshooting.

I would leave tackling the actual issues with the migrations and MariaDB 10.4.31 for others.

juherr:
Ok. It makes sense. But why not trying to fix the root issue instead?

At least, add a todo comment that explains it is a workaround with a link to the related issue issue.

slachiewicz:
MariaDB 10.4.31 Docker image was released 4 days ago and it will break environments for everyone pulling fresh 10.4 tags So it's a rational decision to pin to 10.4.30 for the time being.  

Release notes:
https://mariadb.com/kb/en/mariadb-10-4-31-release-notes/
https://mariadb.com/kb/en/mariadb-10-4-31-changelog/

slachiewicz:
MariaDB 10.4.31 Docker image was released 4 days ago and it will break environments for everyone pulling fresh 10.4 tags So it's a rational decision to pin to 10.4.30 for the time being.  

Release notes:
https://mariadb.com/kb/en/mariadb-10-4-31-release-notes/
https://mariadb.com/kb/en/mariadb-10-4-31-changelog/

goekay:
thanks everyone. i agree with the comments here: the PR is not tackling the root cause, but helps to continue providing a working env for builds. this is good enough.

goekay:
obviously, editing docker-compose is not the whole story. our actions are failing due to the same issue. pls see my branch https://github.com/steve-community/steve/tree/experiment_mariadb and the pinned versions in github workflow. i would be happy if you include them in your PR as well.

slachiewicz (approved):

juherr:
If the issue comes from the migration the only solution will be to remove them.

The v1.0 baseline from #1140 can be used to check it.
It will definitively remove the need of extra rights at the same time.

The huge problem is that it will break all existing instances.
But that's just another argument for multiple flavour releases.

@goekay wdyt?

juherr:
@sirhcel thanks for the todo comment! 🙏

PR #1216, opened 2023-08-20
Fix migrations to work with latest MariaDB patch versions

slachiewicz:
see PR #1213

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ jobs:
       matrix:
         os: [ ubuntu-20.04, ubuntu-22.04 ]
         java: [ '11', '17' ]
-        db: [ 'mysql:5.7', 'mysql:8.0', 'mariadb:10.3', 'mariadb:10.4', 'mariadb:10.5', 'mariadb:10.6' ]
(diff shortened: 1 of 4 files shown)


goekay:
you are rewriting history (i.e. modifying executed migrations). this is always problematic due to existing installations and users of steve. i (we) try to make upgrades between versions seamless.

slachiewicz:
Yes I've made movements only locally to files.
This one where I've dropped and recreated FK are equally to FC existing in version before this file. So this change should be transparent I hope.

If not this way, what other options do we have?

juherr:
@slachiewicz 
 
It is not possible to modify migration files because the checksum of the file is stored and any modification will be considered as corruption and will break the next migrations.

> If not this way, what other options do we have?

See my proposition 
https://github.com/steve-community/steve/pull/1213#issuecomment-1685368887_

goekay:
> It is not possible to modify migration files because the checksum of the file is stored and any modification will be considered as corruption and will break the next migration.

exactly. flyway keeps migration executions in a database table. by the way, there is a way to fix checksums and we used it once: https://github.com/steve-community/steve/wiki/FAQ#how-can-i-upgrade-from-1xx-to-2xx but i am hesitant when it comes to this. it is an annoyance for users and we should use this card really carefully and sparingly.

PR #1226, opened 2023-08-29
Prefer constructor injection

juherr:
First baby step for #1202


slachiewicz (approved):

juherr:
Ping @goekay

juherr:
Ping @goekay Should I close or update this PR?

juherr:
> The Spring team generally advocates constructor injection, as it lets you implement application components as immutable objects and ensures that required dependencies are not null. Furthermore, constructor-injected components are always returned to the client (calling) code in a fully initialized state. As a side note, a large number of constructor arguments is a bad code smell, implying that the class likely has too many responsibilities and should be refactored to better address proper separation of concerns.

(https://docs.spring.io/spring-framework/reference/core/beans/dependencies/factory-collaborators.html)

juherr:
Closed because the issue is outdated

src/main/java/de/rwth/idsg/steve/web/controller/AjaxCallController.java
```diff
@@ -51,14 +49,20 @@
...
+    public AjaxCallController(
+            ChargePointRepository chargePointRepository,
+            TransactionRepository transactionRepository,
+            ReservationRepository reservationRepository
+    ) {
+        this.chargePointRepository = chargePointRepository;
+        this.transactionRepository = transactionRepository;
+        this.reservationRepository = reservationRepository;
         objectMapper = new ObjectMapper();
         objectMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
```
goekay, line 67: coming from [here](https://github.com/steve-community/steve/pull/1817#issuecomment-3298337373) as requested.

i want to only comment on this AjaxCallController refactor as an example to illustrate a general way of approaching this:

for me, personally and subjectively, the following kind of constructor injection with `@RequiredArgsConstructor` the best possible way:

Screenshot 2025-09-16 at 14 51 27

the constructor is not doing any initialization other than injecting the dependencies. therefore, it does not have to literally exist in source code and lombok will play along. the result is very compact and there is no noise at all (which i am striving for). 

whenever the constructor is doing _something_ besides injecting, i start to scratch my head because the desired state above cannot be achieved easily... and look for alternative solutions to approximate this.

in this particular case of AjaxCallController, i opted for [this approach](https://github.com/steve-community/steve/pull/1817/files#diff-6fc064c1d6351e1155182c47065c34041903e1468464484d3d65c54a64944ab0):

Screenshot 2025-09-16 at 15 05 14

`createMapper()` is a small helper method. the result is better than forcing the whole constructor to exist in source code IMO.

juherr: Thanks for your feedback.

I agree with your point of view. I also tend to create factory methods when it feels relevant, although it’s debatable when it’s only about two lines.  

What I was aiming to remove were the `@PostConstruct` annotations that initialize the object state instead of relying on the constructor. But it looks like we share the same view here.  

Regarding the mapper in `AjaxCallController`, I now configure it once at the project level and then inject it:  
https://github.com/juherr/evolve/blob/e7bd7fa19f9a168ca925b110520739857fb0947a/steve-ui-jsp/src/main/java/de/rwth/idsg/steve/web/controller/AjaxCallController.java#L53  
(Of course, OCPP has its own dedicated `ObjectMapper`.)

goekay: > I agree with your point of view. I also tend to create factory methods when it feels relevant, although it’s debatable when it’s only about two lines.

sure, but here is the nuance: creating this factory method allows to get rid of the constructor in source code, since lombok's annotation can be applied.

being able to use lombok is the main objective, creating the factory method is the means to end. i have no specific feelings about the means to end. i try to do whatever possible to achieve the main objective.

juherr: Well, you know, Lombok is already a pretty debatable topic on its own (just not my fight) ;)

PR #1229, opened 2023-08-30
Use Spotless

juherr:


juherr:
@slachiewicz I used the Google format by default but I saw you contributed to palantir. I've never use it. Do you have any feedback to share?

juherr:
Ping @goekay WDYT?

@slachiewicz Any advice between Google format and Palantir format?

slachiewicz:
We use the Spotless plugin to format Maven projects. 
https://github.com/apache/maven-parent/blob/maven-parent-40/pom.xml#L1168-L1211

our configuration files can be found here: https://github.com/apache/maven-shared-resources/tree/master/src/main/resources/config

It works with sometimes minor issues like duplicate license headers for the first reformat, and issues with Java 21.
But in general, I would recommend using Spotless - we previously used only the checkstyle plugin and that was a nightmare to fix formatting errors.

slachiewicz:
and we disabled completely checkstyle plugin completely because with Spotless it is redundant.

juherr:
Done on https://github.com/juherr/evolve

PR #1251, opened 2023-09-25
Update OcppTagsRestController.java

agent87:
Correction on the post mapping where success code is 201 instead of 200


juherr (changes requested):
You just updated the swagger documentation and it won't respect the value provided by the spring annotation.

What are you trying to fix?

PR #1322, opened 2023-12-05
Fix issue related to invalid characters in tags being submitted.

friedkiwi:
The default behaviour of SteVe is to create a new OCPP tag if the charging point submits an unknown tag. If this tag contains invalid characters, the entry ends up in an invalid state:

- you cannot see it in the OCPP tag overview page (and also isn't displayed in the unknown tag section)
- if you go to the active/past sessions pane, and click through to the tag over there, you cannot edit or delete the tag since it doesn't pass validation.

This bug gets triggered because some charge points send arbitrary text as a tag when FreeCharge mode is enabled. The Webasto Next (https://charging.webasto.com/int/products/webasto-next/) chargepoint I use triggers this bug, since it uses the #FreeCharging tag ID when freecharging is enabled.

This PR adds the '#' sign to the whitelist of characters allowed, fixing this issue.

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/web/validation/IdTag.java
+++ b/src/main/java/de/rwth/idsg/steve/web/validation/IdTag.java
@@ -36,7 +36,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 @Constraint(validatedBy = {IdTagValidator.class, IdTagListValidator.class})
 public @interface IdTag {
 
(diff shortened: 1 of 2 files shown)


goekay:
thank you @friedkiwi !

friedkiwi:
> @friedkiwi could you pls add a test case in IdTagValidatorTest that addresses this exact use case?

Done.

goekay:
thank you!

PR #1379, opened 2024-02-03
Propose Docker Compose V2 for bringing up containers

sirhcel:
Attempting this with Docker Compose V1 (docker-compose in its latest release 1.29.2) results in dockerize (the application container entry point) failing to resolve the DB host. This gives gives the error message
```
Waiting for tcp://mariadb:3306: dial tcp: lookup mariadb on 127.0.0.11:53: server misbehaving.
```
in the log of the application container makes it exit after a while.

Bringing up the containers with
```
$ docker compose up -d
```
works like a treat (with docker 25.0.2).

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,13 +96,13 @@ SteVe is designed to run standalone, a java servlet container / web server (e.g.
 
 # Docker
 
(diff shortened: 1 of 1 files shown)


goekay:
thanks @sirhcel !

PR #1426, opened 2024-03-29
Fix transaction overview page to list all txns instead of just active.

friedkiwi:
The default view on the Data Management > Transactions page is to only list active transactions. This is annoying, since you most likely want to get an overview of all transactions registered if you go into this view since there is a separate button to only show the active transactions. The code also suggests it was supposed to be the case that all transactions (and not just the active ones) would be shown.

This PR fixes this.

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/web/dto/TransactionQueryForm.java
+++ b/src/main/java/de/rwth/idsg/steve/web/dto/TransactionQueryForm.java
@@ -44,7 +44,7 @@ public class TransactionQueryForm extends QueryForm {
     private boolean returnCSV = false;
 
     @ApiModelProperty(value = "Return active or all transactions? Defaults to ALL")
-    private QueryType type = QueryType.ACTIVE;
+    private QueryType type = QueryType.ALL;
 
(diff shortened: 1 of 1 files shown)


goekay:
> The code also suggests it was supposed to be the case that all transactions (and not just the active ones) would be shown.

i guess you are referring to:

> @ApiModelProperty(value = "Return active or all transactions? Defaults to ALL")

this documentation is in relation to the API, because

>     public static class ForApi extends TransactionQueryForm {
> 
>         public ForApi() {
>             super();
>             setType(QueryType.ALL);
>             setPeriodType(QueryPeriodType.ALL);
>         }
>     }

where we indeed return all. keep in mind: swagger annotations are for the API. this documentation of the field is, hence, for the API.

i would defend rendering just active transactions for this page is better, because:

- the history of all transactions will only get bigger/longer. this would create additional unnecessary overhead on DB.
- is someone (i.e. a normal user) really interested in all transactions, when she goes to this page? or: recent aka active? i would argue for active.
- getting ALL transactions is just a click (or two) away. is this really that annoying?

goekay:
closing this PR due to inactivity.

PR #1394, opened 2024-02-23
migration bug fixed V0_7_7__update.sql

mahmudarslan:
release version : **steve-3.6.0**

`docker-compose up`
steve-migration-7

2024-02-23 17:31:30 [INFO] BUILD FAILURE
2024-02-23 17:31:30 [INFO] ------------------------------------------------------------------------
2024-02-23 17:31:30 [INFO] Total time:  56.074 s
2024-02-23 17:31:30 [INFO] Finished at: 2024-02-23T14:31:30Z
2024-02-23 17:31:30 [INFO] ------------------------------------------------------------------------
2024-02-23 17:31:30 [WARNING] 
2024-02-23 17:31:30 [WARNING] Plugin validation issues were detected in 2 plugin(s)
2024-02-23 17:31:30 [WARNING] 
2024-02-23 17:31:30 [WARNING]  * pl.project13.maven:git-commit-id-plugin:4.9.10
2024-02-23 17:31:30 [WARNING]  * org.flywaydb:flyway-maven-plugin:7.15.0
2024-02-23 17:31:30 [WARNING] 
2024-02-23 17:31:30 [WARNING] For more or less details, use 'maven.plugin.validation' property with one of the values (case insensitive): [BRIEF, DEFAULT, VERBOSE]
2024-02-23 17:31:30 [WARNING] 
2024-02-23 17:31:30 [ERROR] Failed to execute goal org.flywaydb:flyway-maven-plugin:7.15.0:migrate (default) on project steve: org.flywaydb.core.internal.command.DbMigrate$FlywayMigrateException: Migration V0_7_7__update.sql failed
2024-02-23 17:31:30 [ERROR] -----------------------------------
2024-02-23 17:31:30 [ERROR] SQL State  : HY000
2024-02-23 17:31:30 [ERROR] Error Code : 1833
2024-02-23 17:31:30 [ERROR] Message    : Cannot change column 'chargeBoxId': used in a foreign key constraint 'stevedb/FK_chargeBoxId_c' of table 'stevedb/connector'
2024-02-23 17:31:30 [ERROR] Location   : /code/src/main/resources/db/migration/V0_7_7__update.sql (/code/src/main/resources/db/migration/V0_7_7__update.sql)
2024-02-23 17:31:30 [ERROR] Line       : 1
2024-02-23 17:31:30 [ERROR] Statement  : ALTER TABLE `chargebox`
[...]

PR #1428, opened 2024-03-31
migration bug fixed  V0_8_6__update.sql

mahmudarslan:


juherr:
You cannot update migrations that were already released because it will broke servers currently in production: an hash of the migration file is done and store in database, and will fail if a previous migration changed.

juherr:
Related to #1394

juherr:
Should be fixed by #1455

goekay:
what is the state of this PR after merging #1439 ?

goekay:
i am closing this issue due to inactivity. feel free to reopen if it is still relevant.

Issue #1417, opened 2024-03-23
Can't compile SteVe 3.6.0


d6e6t6a6l6e6s:
**Hello, I installed a new server and I can't compile SteVe 3.6.0. What happened ? Looks like I do the same thing...**

SteVe Version     : SteVe 3.6.0
Operating system  : Debian 11 bullseye
JDK  : openjdk version "11.0.22" 2024-01-16
Database : 10.5.23-MariaDB-0+deb11u1 Debian 11

```
#  ./mvnw package
Warning: JAVA_HOME environment variable is not set.
[INFO] Scanning for projects...
[INFO]
[INFO] -------------------------< de.rwth.idsg:steve >-------------------------
[INFO] Building SteVe 3.6.0
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- enforcer:3.3.0:enforce (enforce-java) @ steve ---
[INFO] Rule 0: org.apache.maven.enforcer.rules.version.RequireJavaVersion passed
[INFO]
[INFO] --- git-commit-id:4.9.10:revision (default) @ steve ---
[INFO]
[INFO] --- properties:1.1.0:read-project-properties (default) @ steve ---
[INFO]
[INFO] --- flyway:7.15.0:migrate (default) @ steve ---
[INFO] Flyway Community Edition 7.15.0 by Redgate
[INFO] Database: jdbc:mysql://localhost:3306/stevedb (MySQL 5.5)
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:10 min
[INFO] Finished at: 2024-03-23T10:10:57Z
[INFO] ------------------------------------------------------------------------
[WARNING]
[WARNING] Plugin validation issues were detected in 2 plugin(s)
[WARNING]
[WARNING]  * pl.project13.maven:git-commit-id-plugin:4.9.10
[WARNING]  * org.flywaydb:flyway-maven-plugin:7.15.0
[WARNING]
[WARNING] For more or less details, use 'maven.plugin.validation' property with one of the values (case insensitive): [BRIEF, DEFAULT, VERBOSE]
[WARNING]
[ERROR] Failed to execute goal org.flywaydb:flyway-maven-plugin:7.15.0:migrate (default) on project steve: org.flywaydb.core.api.exception.Flyw                                      ayValidateException: Validate failed: Migrations have failed validation
[...]

PR #1439, opened 2024-04-20
Setting a database baseline

fnkbsi:
Some combination of DB versions and OS versions seems to have problems with the DB migration scripts. 
@juherr suggested (https://github.com/steve-community/steve/pull/1394#issuecomment-2031314233) a baseline script to solve the issues (e.g. #1417). 
The PR #1394 and #1428   addresses the same issues, but changes the existing migration scripts, which could causes trouble at updating existing Steve servers. 

The baseline also improved slightly the build process, because less DB build/migration operation are executed.

The script is exported (by mariadb-dump) from a fresh build Steve instance. The baseline creates all tables and views and inserts the data of the setting table.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,7 +35,6 @@ jobs:
           mysql -h 127.0.0.1 -P 3306 -uroot -proot -e "CREATE DATABASE stevedb_test_2aa6a783d47d;" -v
           mysql -h 127.0.0.1 -P 3306 -uroot -proot -e "CREATE USER 'steve'@'%' IDENTIFIED BY 'changeme';" -v
           mysql -h 127.0.0.1 -P 3306 -uroot -proot -e "GRANT ALL PRIVILEGES ON stevedb_test_2aa6a783d47d.* TO 'steve'@'%';" -v
-          mysql -h 127.0.0.1 -P 3306 -uroot -proot -e "GRANT SELECT ON mysql.proc TO 'steve'@'%';" -v || true
           mysql -h 127.0.0.1 -P 3306 -uroot -proot -e "GRANT SUPER ON *.* TO 'steve'@'%';" -v || true
 
       - name: Build with Maven
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,7 +35,6 @@ jobs:
           mysql -h 127.0.0.1 -P 3306 -uroot -proot -e "CREATE DATABASE stevedb_test_2aa6a783d47d;" -v
           mysql -h 127.0.0.1 -P 3306 -uroot -proot -e "CREATE USER 'steve'@'%' IDENTIFIED BY 'changeme';" -v
           mysql -h 127.0.0.1 -P 3306 -uroot -proot -e "GRANT ALL PRIVILEGES ON stevedb_test_2aa6a783d47d.* TO 'steve'@'%';" -v
-          mysql -h 127.0.0.1 -P 3306 -uroot -proot -e "GRANT SUPER ON *.* TO 'steve'@'%';" -v || true
 
       - name: Build with Maven
         run: ./mvnw -B -V -Dmaven.javadoc.skip=true -Ptest clean package --file pom.xml
--- a/src/main/resources/db/migration/B1_0_5__stevedb.sql
+++ b/src/main/resources/db/migration/B1_0_5__stevedb.sql
@@ -319,6 +319,12 @@ CREATE TABLE `settings` (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_unicode_ci;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
+
+-- Exported data of table stevedb.settings: ~1 rows (approx.)
+INSERT INTO `settings` (`app_id`, `heartbeat_interval_in_seconds`, `hours_to_expire`, `mail_enabled`, `mail_host`, `mail_username`, `mail_password`, `mail_from`, `mail_protocol`, `mail_port`, `mail_recipients`, `notification_features`) VALUES
+	('U3RlY2tkb3NlblZlcndhbHR1bmc=', 14400, 1, 0, NULL, NULL, NULL, NULL, 'smtp', 25, NULL, NULL);
+
+
 --
 -- Temporary table structure for view `transaction`
 --
--- a/src/main/resources/db/migration/B1_0_5__stevedb.sql
+++ b/src/main/resources/db/migration/B1_0_5__stevedb.sql
@@ -5,13 +5,13 @@
 -- ------------------------------------------------------
 -- Server version	11.4.2-MariaDB
 
-/*!40101 SET NAMES utf8mb4 */;
-/*!40103 SET TIME_ZONE='+00:00' */;
-/*!40101 SET character_set_client = utf8 */;
+SET NAMES utf8mb4;
+SET TIME_ZONE='+00:00';
+SET character_set_client = utf8;
 
-/*!40014 SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0 */;
-/*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
-/*!40101 SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='NO_AUTO_VALUE_ON_ZERO' */;
+SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0;
+SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0;
+SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='NO_AUTO_VALUE_ON_ZERO';
 
 --
 -- Table structure for table `address`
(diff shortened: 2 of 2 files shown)


juherr:
That's nice but, as I know, it won't work because baseline need at least Flyway Teams 😓
https://documentation.red-gate.com/fd/feature-glossary-165740620.html

fnkbsi:
> That's nice but, as I know, it won't work because baseline need at least Flyway Teams 😓 https://documentation.red-gate.com/fd/feature-glossary-165740620.html

I tried it and it worked (with a additional migration script). So I hope maven use the Flyway baseline (command) which is included in the Flyway Community version. 
Can someone confirm?

juherr:
I didn't notice baseline command is available, thanks! I will check asap

juherr:
@fnkbsi I didn't check yet but if the baseline is working, you should be able to remove this lines without breaking the CI https://github.com/steve-community/steve/blob/master/.github/workflows/main.yml#L37-L39

Could you try?

juherr:
@fnkbsi For historical reason, I think the sources should include [the V1_0_0 baseline](https://github.com/steve-community/steve/pull/1140/files#diff-647a1a98ffe4ec0fb384126b0a3518fb0ff3607e22d0df116212550e90b6d908). WDYT?

fnkbsi:
> @fnkbsi I didn't check yet but if the baseline is working, you should be able to remove [this lines](https://github.com/steve-community/steve/blob/master/.github/workflows/main.yml#L37-L39) without breaking the CI.
> 
> Could you try?

Why should we change the privileges at the workflow script? There is no difference between the update and baseline scripts, the baseline scripts just replaces the update scripts up to the correspondent version number on a new installation.

fnkbsi:
> @fnkbsi For historical reason, I think the sources should include [the V1_0_0 baseline](https://github.com/steve-community/steve/pull/1140/files#diff-647a1a98ffe4ec0fb384126b0a3518fb0ff3607e22d0df116212550e90b6d908). WDYT?

Setting the baseline at version 1_0_0 is also possible, just renaming the script V1_0_0 baseline.sql to B1_0_0 baseline.sql should do the trick.
Two baseline scripts in this state of development are not necessary, so @goekay should decide which version becomes be the baseline.

juherr:
> Why should we change the privileges at the workflow script? There is no difference between the update and baseline scripts, the baseline scripts just replaces the update scripts up to the correspondent version number on a new installation.

Because the extra rights were due to migration before B1_0_0.
With B1_0_0 we don't need the extra rights anymore.

> just renaming the script V1_0_0 baseline.sql to B1_0_0 baseline.sql should do the trick.

That was my point, could you add B1_0_0 too?

fnkbsi:
> > just renaming the script V1_0_0 baseline.sql to B1_0_0 baseline.sql should do the trick.
> 
> That was my point, could you add B1_0_0 too?

If I add the B1_0_0 and don't remove the B1_0_5 the B1_0_0 script is dead code, because it will never be used. So I am hesitant to integrated it only for cosmetics.

goekay:
hey there, i am up to speed with this change and discussion. sorry for the delay.

regarding where to set the baseline:
- current PR takes the latest snapshot of DB and makes it a baseline. with this approach, for fresh installations, the only db migration will be B1_0_5.
- @juherr was suggesting, i think, setting the baseline to an earlier snapshot after which versioned migrations should continue as usual. with this approach, for fresh installations, the migration sequence will be: B1_0_0, V1_0_1, V1_0_2, V1_0_3, V1_0_4, V1_0_5. afaik, he was not suggesting to keep both baselines.

i am actually fine with both. i kinda like @juherr's suggestion a bit better because granularity of changes will be kept instead of one bulk monolith.

my concern is these exports coming from heidi sql. i am not sure how it works, but i would rather prefer mysql's (or mariadb's) native dump functionality to prevent man-in-the-middle interpretations and probable errors. is this possible? or, is my concern even valid? wdyt?

talking about mysql and mariadb... i see that the matrix builds were fine but, this baseline will be compatible with both, right?  nuances started to appear in new mariadb versions because of which mariadb started to drift from mysql.

juherr:
Yes. I think major versions are better candidate for baselines.

I've proposed an alternative pr #1455 which removes mysql privileges that are useless after v1.0

goekay:
> I think major versions are better candidate for baselines.

i was not following semantic versioning with db migration files though. these are just linear changes to be applied. 

my approach was: 
- bump the version of the lowest level every time you make a change (as long as the versions remain in single digits).
- if the previous step is about the cause double digits, bump the higher level's version.

example 1 : 0.8.8 -> 0.8.9 (fine)
example 2: 0.8.9 -> 0.8.10 (uh oh) -> 0.9.0 (fine again)
example 2: 1.1.9 -> 1.1.10 (uh oh) -> 1.2.0 (fine again)

juherr:
Funny versionning convention. I didn't get that before. Thanks for the clarification.

In that case whatever version is used as baseline will be ok.
In my opinion, the best part of using a baseline is the drop of extra rights on the database.

About my PR, I made the baseline by hand and applied every fixes myself to be sure to respect the history.

fnkbsi:
> i am actually fine with both. i kinda like @juherr's suggestion a bit better because granularity of changes will be kept instead of one bulk monolith.
> 
In general I agree, but  V1_0_1 - V1_0_4 include only 'Alter Table' commands with minor changes. 
1_0_5 added two views, so the length of the script is slightly longer.

> my concern is these exports coming from heidi sql. i am not sure how it works, but i would rather prefer mysql's (or mariadb's) native dump functionality to prevent man-in-the-middle interpretations and probable errors. is this possible? or, is my concern even valid? wdyt?
> 
There are differences in the exported/dumped scripts. On a first glance I've seen two or three general differences (except the comments).
The first difference is maria-dump drops the tables and creates a new, the heidi script creates the table only if the table not exists. Then there are differences on creating temporary tables and in the executable comments. These differences I haven't further investigated yet.

> talking about mysql and mariadb... i see that the matrix builds were fine but, this baseline will be compatible with both, right? nuances started to appear in new mariadb versions because of which mariadb started to drift from mysql.
>
Based on the github actions results, yes. Also I tested it locally with a mariadb v11 on my windows machine.

leomwa:
FWIW, I pulled down this branch and attempted a docker compose up, but it failed:
```
2024-06-08 06:19:19 [INFO] Rule 1: org.apache.maven.enforcer.rules.version.RequireMavenVersion passed
2024-06-08 06:19:19 [INFO] ------------------------------------------------------------------------
2024-06-08 06:19:19 [INFO] BUILD FAILURE
2024-06-08 06:19:19 [INFO] ------------------------------------------------------------------------
2024-06-08 06:19:19 [INFO] Total time:  8.550 s
2024-06-08 06:19:19 [INFO] Finished at: 2024-06-08T04:19:19Z
2024-06-08 06:19:19 [INFO] ------------------------------------------------------------------------
2024-06-08 06:19:19 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.4.1:enforce (enforce-java) on project steve: 
2024-06-08 06:19:19 [ERROR] Rule 0: org.apache.maven.enforcer.rules.version.RequireJavaVersion failed with message:
2024-06-08 06:19:19 [ERROR] Detected JDK version 11.0.20 (JAVA_HOME=/opt/java/openjdk) is not in the allowed range [17,).
2024-06-08 06:19:19 [ERROR] -> [Help 1]
2024-06-08 06:19:19 [ERROR] 
2024-06-08 06:19:19 [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
2024-06-08 06:19:19 [ERROR] Re-run Maven using the -X switch to enable full debug logging.
2024-06-08 06:19:19 [ERROR] 
2024-06-08 06:19:19 [ERROR] For more information about the errors and possible solutions, please read the following articles:
2024-06-08 06:19:19 [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

goekay:
if @leomwa made a fresh pull of the branch without modifications, it should be java 17 already. Dockerfile we use references java 17. therefore, i am confused since i dont know where this java 11 comes into play.

goekay:
@fnkbsi i see you made some changes after my comment. are these changes in reaction to my remarks? if yes, are you finished? is this PR stable now?

fnkbsi:
@goekay  Yes after your remarks I changed the base of the script on the mysqldump.exe script. The PR is stable unless there are more comments.

goekay:
thanks! migration file LGTM.

i think you can even remove the `GRANT SUPER ...` privileges from CI and from the [readme here](https://github.com/steve-community/steve?tab=readme-ov-file#configuration-and-installation).

fnkbsi:
> i think you can even remove the `GRANT SUPER ...` privileges from CI and from the [readme here](https://github.com/steve-community/steve?tab=readme-ov-file#configuration-and-installation).

Without the Super privileges the workflow does not run successful. MySql need SET_USER_ID privilege (https://github.com/fnkbsi/steve/actions/runs/9641524165) and mariadb struggles with the views without the Super privilege (https://github.com/fnkbsi/steve/actions/runs/9641825097).

goekay:
> Without the Super privileges the workflow does not run successful.

... because the view creation dictates `steve@localhost` as definer which is not necessarily needed. 

- see [this](https://stackoverflow.com/questions/13492857/create-view-without-super-privileges-in-phpmyadmin)
- [my change](https://github.com/steve-community/steve/commit/7d13dbf22190267c16ada73cf907055600b5b967) in a test branch (forked from your PR branch) and [action results](https://github.com/steve-community/steve/actions/runs/9644220689)

goekay (approved):

goekay:
@fnkbsi should i merge or do you want to? i approved the PR to signal that you can merge it.

fnkbsi:
@goekay: Please merge.

goekay:
thanks all!

PR #1460, opened 2024-05-24
Feature/not fail on receiving with unknown ocpp properties

dakai-wei-of-shizen:
we are trying to use SteVe to provide customer the OCPP Server, facing these problems and solving it. 
* some CP added custom ocpp fields(properties) in their request from CP or response, mostly for debug or exception handling, making the steve ocpp server not working as expected.  
    =>So I made it not to fail on receiving with unknown ocpp properties
* _(implemented BUT REVERSED)some of our customers' CP can only set IP and Port in their CP, making the websocket endpoint not accessible .
    =>So I added  a reverse proxy to handle this locally first.
    =>Not related to the title of the Pull Request so it is REVERSED._


goekay:
thanks for the contribution. can you pls write a description and motivation for the changes? why are you adding nginx? this PR is mixing two concerns.

dakai-wei-of-shizen:
> thanks for the contribution. can you pls write a description and motivation for the changes? why are you adding nginx? this PR is mixing two concerns.

Thanks for your remind!
I didn't intend to open a PR in this origin repo, the diff is only for dev local env, I can close it. Sorry for the disturbing.

goekay:
hey there, thanks for the updates and motivation.

i find both of these optional, but i like one better than the other.

- nginx: this is too close to infra decisions. as you can see, the codebase of steve does not tell/dictate anything how your infra should look like and how you should install steve. we leave infra-relation decisions to the user, because there are many variants. makes sense?
- custom msg fields: if they add custom fields to messages, this would mean that they do NOT create ocpp-compliant messages anymore. this is bad for the standard. "strict me" would say no, but i can understand where it is coming from. since this is not a destructional change, i am okay with having this.

dakai-wei-of-shizen:
@goekay 
> nginx: 

Got it. 

> custom msg fields:

Understood. I will make it configureable in the main.properties and recreate the PR. 

Closing this PR.

goekay:
i am actually fine with the current state of custom msg fields. no need for additional configuration.

i just would like to omit the nginx changes.

dakai-wei-of-shizen:
@goekay 
As you wish!
Made it ready for review.

Issue #1473, opened 2024-06-03
Connector 0/zero causes Exception


andle:
#### Checklist

- [ x] I checked other issues already, but found no answer/solution
- [ x] I checked the documentation and wiki, but found no answer/solution
- [ x] I am running the latest version and the issue still occurs
- [ x] I am sure that this issue is about SteVe (and not about the charging station software or something unrelated to SteVe)

#### Specifications
```
SteVe Version     : 3.6.0
Operating system  : Linux 5.4.0-1108-raspi #120-Ubuntu
JDK               : openjdk 17.0.10 2024-01-16
                        OpenJDK Runtime Environment (build 17.0.10+7-Ubuntu-120.04.1)
                        OpenJDK 64-Bit Server VM (build 17.0.10+7-Ubuntu-120.04.1, mixed mode, sharing)

Database          : mariadb  Ver 15.1 Distrib 10.3.39-MariaDB, for debian-linux-gnu (aarch64) using readline 5.2
```

#### Expected Behavior

Send a ChargePointMaxProfile with connector set to 0 with no errors.

#### Actual Behavior

Charge profile is Accepted, but Java errors occur due to connector being zero/"null".

#### Steps to Reproduce the Problem

  1. Create a ChargePointMaxProfile (any valid Max profile seems to do)
  2. Set Charging Profile > select charger and profile > Set connector ID to 0 (zero)
  3. Click "Perform"
  4. Observe steve.log to see exception/error

#### Additional context

From steve.log:
[INFO ] 2024-06-03 17:35:27,793 de.rwth.idsg.steve.ocpp.ws.WebSocketLogger (SteVe-Executor-0) - [chargeBoxId=ST19A7114B1EE, sessionId=196ac060-e7ce-4a59-d4d2-0c0f228b7970] Sending: [2,"1dd34e02-476a-429d-b1bf-0d
7e20397ec9","SetChargingProfile",{"connectorId":0,"csChargingProfiles":{"chargingProfileId":1,"stackLevel":1,"chargingProfilePurpose":"ChargePointMaxProfile","chargingProfileKind":"Absolute","chargingSchedule":{
"chargingRateUnit":"A","chargingSchedulePeriod":[{"startPeriod":0,"limit":8,"numberPhases":1}]}}}]                                                                                                                 
[...]

PR #1474, opened 2024-06-03
insert connector conditionally for SetChargingProfile tasks (#1473)

goekay:
if the connector (for which a SetChargingProfile was sent) was not observed this far (i.e. the station did not report it), insert it before saving profile info.

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/repository/impl/ChargingProfileRepositoryImpl.java
+++ b/src/main/java/de/rwth/idsg/steve/repository/impl/ChargingProfileRepositoryImpl.java
@@ -68,6 +68,8 @@ public class ChargingProfileRepositoryImpl implements ChargingProfileRepository 
 
     @Override
     public void setProfile(int chargingProfilePk, String chargeBoxId, int connectorId) {
+        OcppServerRepositoryImpl.insertIgnoreConnector(ctx, chargeBoxId, connectorId);
+
         SelectConditionStep<Record1<Integer>> connectorPkSelect = ctx.select(CONNECTOR.CONNECTOR_PK)
                                                                      .from(CONNECTOR)
                                                                      .where(CONNECTOR.CHARGE_BOX_ID.eq(chargeBoxId))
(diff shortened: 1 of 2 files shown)


fnkbsi:
If I interpret the solution/code correct: A connector ID which is not included in the DB, will be added to the DB, if the connector ID is insert to the form on the "Set Charging Profile" operation site. Even if the connector does not exist? 
That's would be in my Opinion OK for connector ID = 0 (references the whole charge box). For all other connector IDs I would prefer a warning on the "Set Charging Profile" site. Than the User can add the connector manually, if the connector ID exists.
Alternative using a connector ID Drop-Down List on the site.

goekay:
i see your concern. i was thinking as follows: this request first will be sent to the station, which should verify/validate the input and then accept or reject it. if the request contains a connector id that the station does not posses, it should reject the request. we are saving this connector (and the profile) in the database only in cases of acceptance.

therefore, i did not want to introduce a complicated user flow or implementation as you suggest. i can reduce this functionality to only connector id 0 though.

fnkbsi:
> i see your concern. i was thinking as follows: this request first will be sent to the station, which should verify/validate the input and then accept or reject it. if the request contains a connector id that the station does not posses, it should reject the request. we are saving this connector (and the profile) in the database only in cases of acceptance.
> 
> therefore, i did not want to introduce a complicated user flow or implementation as you suggest. 
I like your new solution! 

>i can reduce this functionality to only connector id 0 though.
a) With the new approach it should not be necessary to restrict it to connector id 0.
b) The current approach with the restriction (connector id = 0) is less code/complicated than your new solution.

So for me the questions is: Are there reports of missing connector id's other than 0? 
Depending on the answer my suggestion is:
 If yes -> implementing the new approach without restriction
 If no -> current approach with restriction to ID 0

goekay:
> I like your new solution!

this is not a new solution, this is the current state. i was just giving justification for my changes. if you follow the call to the method i changed, you will see [this](https://github.com/steve-community/steve/blob/52794d15ec8e83c9c3cadc91a0ca6e3236b10508/src/main/java/de/rwth/idsg/steve/ocpp/task/SetChargingProfileTask.java#L62-L65)

goekay:
@fnkbsi wdyt?

fnkbsi:
@goekay: My mistake, I thought the connector check and insert take place before sending the profile. After your clarification there  are no concerns on my side. 

So only if the charge box accepts a profile for an non existing connector we would add this connector to the DB. In that case the error happens on charge box side and Steve is not responsible, neither can Steve verify the connector otherwise.

goekay:
> So only if the charge box accepts a profile for an non existing connector we would add this connector to the DB. In that case the error happens on charge box side and Steve is not responsible, neither can Steve verify the connector otherwise.

exactly. 

alright, with that clarified i am merging the PR.

PR #1511, opened 2024-07-14
better msg if logs are unavailable (closes #1503)

### LINKED ISSUE #1503: Logs are Unavailable
SteVe Version     : 3.6.0
Operating system  : Ubuntu 22.04 Virtual Machine Running on Windows 11 Host
JDK               :  java 21.0.2  on host

#### Expected Behavior

when hitting the log button on the steve web interface, a log will appear

#### Actual Behavior

a mostly blank page appears that says steve log unavailable
not available

#### Steps to Reproduce the Problem

  1. open steve directory
  2. start docker 
  3. open web interface
  4. hit logs button in top right
  5. get logs unavailable error

goekay:
@juherr @lmccarty5 would this be an improvement?

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/utils/LogFileRetriever.java
+++ b/src/main/java/de/rwth/idsg/steve/utils/LogFileRetriever.java
@@ -68,7 +68,7 @@ public enum LogFileRetriever {
     }
 
     public String getErrorMessage() {
-        return "Not available";
+        return "Logs are not available here, because they are being directed to console/stdout instead";
     }
 
     // -------------------------------------------------------------------------


juherr:
@goekay I think it is better but the log section could be hidden when not available too.

goekay:
> the log section could be hidden when not available too.

sure, the way to do is as follows: expose the condition to hide in a model parameter (e.g. the profile steve runs at), and put [this line](https://github.com/steve-community/steve/blob/master/src/main/resources/webapp/WEB-INF/views/00-header.jsp#L73) behind that condition.

but, when switching between profiles the header sequence would not be consistent (sometimes log is there, sometimes not, ...). i would argue that this could cause a confusion as well: why is this happening?

with the solution from this PR, we would be telling _the why_ to the user.

juherr (approved):

PR #1519, opened 2024-07-20
migrate to jakarta and upgrade spring, jetty, cxf

goekay:
we were really behind of latest versions with some of our big dependencies (spring, jetty, cxf). the migration from javax to jakarta was messy for the libraries and frameworks. and i was waiting for all of our downstream dependencies to finish their work, because a partial update would have been problematic (e.g. upgrade spring only which switched to jakarta, but leave jetty on javax).


juherr (approved):

goekay:
the project finally compiles. 

but there are some runtime errors + some code that i commented out because the APIs changed. i will address these cases later.

slachiewicz (approved):

goekay:
dev profile runs without issues. prod has problems with security and openapi.

goekay:
it is working without issues based on my tests. i want to encourage you, the community member, to test yourself as well. 

i will leave this PR open for a couple of days. i will merge it maybe saturday afternoon/evening. before i do it, i will make a new github release to mark the current status as stable, since these big changes/migrations might introduce some problems.

Issue #1526, opened 2024-08-01
Validate chargeBoxId for WS connections


goekay:
#### Checklist

- [x] I checked other issues already, but found no answer/solution
- [x] I checked the documentation and wiki, but found no answer/solution
- [x] I am running the latest version and the issue still occurs
- [x] I am sure that this issue is about SteVe (and not about the charging station software or something unrelated to SteVe)

#### Specifications
```
SteVe Version     : ...
Operating system  : ...
JDK               : ...
Database          : ...
```

#### Expected Behavior

Steve should not allow [any arbitrary last bit of URL](https://github.com/steve-community/steve/blob/master/src/main/java/de/rwth/idsg/steve/ocpp/ws/OcppWebSocketHandshakeHandler.java#L72) to enter the system. We actually have [ChargeBoxIdValidator](https://github.com/steve-community/steve/blob/master/src/main/java/de/rwth/idsg/steve/web/validation/ChargeBoxIdValidator.java) that can be used to make sure that this String is not in a problematic format.

#### Actual Behavior

...

#### Steps to Reproduce the Problem

  1. ...
  2. ...
  3. ...

#### Additional context

...

goekay:
The regex we are using allows any non-whitespace character which is too generous. I am thinking about applying some restrictions similar to [IdTagValidator](https://github.com/steve-community/steve/blob/master/src/main/java/de/rwth/idsg/steve/web/validation/IdTagValidator.java) but without the length constraint. 

wdyt @juherr

juherr:
@goekay I think it should be configurable and disabled by default. 
Otherwise it will break existing instances where blanks are currently used (by choice or error).

In case the option is not enabled, a warn log may be used for each stations registred with unexpected characters.

juherr:
@goekay why not having an option that sanitizes names and remove unexpected chars? It can help to follow recommandations without breaking things.

goekay:
> I think it should be configurable and disabled by default.
> Otherwise it will break existing instances where blanks are currently used (by choice or error).

i try to prevent configuration/property explosion. it would be great if a solution without property is possible.

> why not having an option that sanitizes names and remove unexpected chars? It can help to follow recommandations without breaking things.

because the "input" will differ from the "entry" in our system, and it might lead to confusion.

PR #1527, opened 2024-08-01
validate chargeBoxId for WS connections (#1526)

goekay:


src/main/java/de/rwth/idsg/steve/ocpp/ws/OcppWebSocketHandshakeHandler.java
```diff
@@ -48,6 +49,8 @@
 @RequiredArgsConstructor
 public class OcppWebSocketHandshakeHandler implements HandshakeHandler {
 
+    private final ChargeBoxIdValidator chargeBoxIdValidator = new ChargeBoxIdValidator();
```
juherr, line 52: It could be static

juherr (approved):

src/test/java/de/rwth/idsg/steve/web/validation/ChargeBoxIdValidatorTest.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Test
+    public void testEmpty() {
+        Assertions.assertFalse(validator.isValid("", null));
+    }
+
+    @Test
+    public void testSpace() {
+        Assertions.assertFalse(validator.isValid("  ", null));
```
juherr, line 109: You should test some other blank, not printable, characters from the ascii table which are currently accepted.

src/main/java/de/rwth/idsg/steve/web/validation/ChargeBoxIdValidator.java
```diff
@@ -28,7 +28,7 @@
  */
 public class ChargeBoxIdValidator implements ConstraintValidator<ChargeBoxId, String> {
 
-    private static final String REGEX = "\\S+";
+    private static final String REGEX = "^[a-zA-Z0-9.:_#-]+$";
```
juherr, line 31: It's a bit restrictive. Why not accepting `Â` (I sadly have some stations with it 🥲) and others which are currently accepted?

juherr (commented):
I think the new regex is too restrictive and will break instances.

goekay:
> I think the new regex is too restrictive and will break instances.

fair enough, and i was fearing this as well. 

new suggestion: instead of whitelisting what is allowed in the regex, i could blacklist what should not appear in the regex. i am thinking something like `[^=/()<>]*` i.e. every character with the exception of `=/()<>` is allowed.

juherr:
The blacklist solution sounds better.

But even if I'll be surprised that a charger uses one of this chars somewhere you need a way to block the upgrade if a charger uses one of them in the database.
Otherwise some users will lost the charger without any way to change its configuration.

goekay:
> But even if I'll be surprised that a charger uses one of this chars somewhere you need a way to block the upgrade if a charger uses one of them in the database.
Otherwise some users will lost the charger without any way to change its configuration.

i wonder how big that percentage is. if we are about to lose a minority (for ex 0.1%) of installations, i would be okay to make that sacrifice.

i can put the new logic behind a config, but i then i would like to have it enabled by default. installations that have problems can still opt out, while accepting the risks.

juherr:
It can be enabled by default but I think it must be possible to disable it or block the upgrade if one station doesn't respect the convention.
Otherwise the hardware will be totally lost with no way to restore it.

juherr (commented):
The regex could be improved

src/main/resources/webapp/WEB-INF/views/data-man/chargepoints.jsp
```diff
@@ -48,7 +48,7 @@
             <tbody>
             <c:forEach items="${unknownList}" var="item">
                 <tr>
-                    <td>${item.key}</td>
+                    <td><encode:forHtml value="${item.key}" /></td>
```
juherr, line 51: The same fix should be applied in some other places (not an exhaustive list):
https://github.com/steve-community/steve/blob/master/src/main/resources/webapp/WEB-INF/views/data-man/chargepointDetails.jsp#L89-L93
https://github.com/steve-community/steve/blob/master/src/main/resources/webapp/WEB-INF/views/data-man/chargepointDetails.jsp#L95-L96
https://github.com/steve-community/steve/blob/master/src/main/resources/webapp/WEB-INF/views/data-man/transactions.jsp#L122
https://github.com/steve-community/steve/blob/master/src/main/resources/webapp/WEB-INF/views/data-man/transactionDetails.jsp#L39
https://github.com/steve-community/steve/blob/master/src/main/resources/webapp/WEB-INF/views/data-man/transactionDetails.jsp#L62
https://github.com/steve-community/steve/blob/master/src/main/resources/webapp/WEB-INF/views/data-man/ocppTags.jsp#L51

src/main/java/de/rwth/idsg/steve/web/validation/ChargeBoxIdValidator.java
```diff
@@ -40,4 +43,22 @@ public void initialize(ChargeBoxId idTag) {
...
         return string == null || PATTERN.matcher(string).matches();
     }
+
+    public boolean isValid(String str) {
+        if (Strings.isNullOrEmpty(str)) {
+            return false;
+        }
+
+        String str1 = str.strip();
+        if (!str1.equals(str)) {
```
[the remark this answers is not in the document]
goekay: your new version is the demonstration of what i was trying to prevent: an unreadable complexity hard to understand.

dont get me wrong. regex is a powerful machine, but too much of it can cause head aches. therefore, i try to minimize the use of it and KISS.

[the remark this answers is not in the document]
goekay: maybe you wanted to give this decision (whether spaces at at start and end are allowed or not) to the user, by making this a part of the regex which can be overridden?

if yes: i think i am making a reasonable opinionated decision here by not allowing them.

if no: then, we are just talking about refactor. i argue that simple java code is simpler than complex regex.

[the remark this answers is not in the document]
juherr: My point is still the same: if it is not configurable, it will be dangerous for some users and you may lost some chargers without any simple way to restore them.

I agree that regex may be sometimes difficult to read but you rarely do it and tests are here to validate the expected behavior.

I think the tradeoff is acceptable here.

Otherwise, another configuration is possible too.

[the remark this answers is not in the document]
juherr: I missed that spaces were not allowed before then forget my point.

juherr (approved):
LGTM except the ui fix which should be extended everywhere a displayed data comes from a station (maybe in another pr?).

goekay:
yes, there will be another issue and another PR.

src/main/resources/webapp/WEB-INF/views/data-man/chargepoints.jsp
goekay: see https://github.com/steve-community/steve/pull/1527#issuecomment-2267115652

PR #1534, opened 2024-08-08
fix ObjectMapper used for API endpoint errors

goekay:
reason: warnings like the following

> [WARN ] 2024-08-08 23:34:20,844 org.eclipse.jetty.ee10.servlet.ServletChannel (qtp739264372-28) - handleException /steve/api/v1/transactions com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Joda date/time type `org.joda.time.DateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-joda" to enable handling (through reference chain: de.rwth.idsg.steve.web.api.ApiControllerAdvice$ApiErrorResponse["timestamp"])

ApiDocsConfiguration activates JacksonAutoConfiguration which creates a default/primary ObjectMapper
that is different from our ObjectMapper. this came with the spring 6.x migration since OpenApi integration
was massively refactored with that as well.

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/config/BeanConfiguration.java
+++ b/src/main/java/de/rwth/idsg/steve/config/BeanConfiguration.java
@@ -21,7 +21,6 @@ package de.rwth.idsg.steve.config;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
-import com.fasterxml.jackson.datatype.joda.JodaModule;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.mysql.cj.conf.PropertyKey;
 import com.zaxxer.hikari.HikariConfig;
@@ -39,7 +38,6 @@ import org.jooq.conf.Settings;
 import org.jooq.impl.DSL;
 import org.jooq.impl.DataSourceConnectionProvider;
 import org.jooq.impl.DefaultConfiguration;
-import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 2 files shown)


src/main/java/de/rwth/idsg/steve/config/BeanConfiguration.java
```diff
@@ -266,6 +267,7 @@ public void extendMessageConverters(List<HttpMessageConverter<?>> converters) {
      * {@link WebMvcConfigurationSupport#requestMappingHandlerAdapter(ContentNegotiationManager, FormattingConversionService, org.springframework.validation.Validator)}.
      */
     @Bean
+    @Qualifier("steveObjectMapper")
```
juherr, line 270: As you only use it in security config, maybe you should move it security config too.

goekay: i prefer to have this objectmapper in close proximity to `extendMessageConverters(..) ` where i modify and configure it.

juherr:
The default bean name is `jacksonObjectMapper`
https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java#L130

Just rename and it should work as previously:

```diff
- public ObjectMapper objectMapper(RequestMappingHandlerAdapter requestMappingHandlerAdapter)
+ public ObjectMapper jacksonObjectMapper(RequestMappingHandlerAdapter requestMappingHandlerAdapter)
```

But maybe the issue is just using `JacksonAutoConfiguration` (a spring boot dependency btw). What were the missing beans when you added springdoc?

goekay:
> But maybe the issue is just using JacksonAutoConfiguration (a spring boot dependency btw). What were the missing beans when you added springdoc?

the swagger ui needs some elements of provided by JacksonAutoConfiguration. it is not just ObjectMapper. when i remove JacksonAutoConfiguration, the swagger ui just does not load... at all.

PR #1539, opened 2024-08-11
Implement database-based multi user system for Web UI

goekay:
my take on #991 to be compared with #1165

FYI @fnkbsi @juherr

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
+++ b/src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
@@ -20,6 +20,7 @@ package de.rwth.idsg.steve.config;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
+import de.rwth.idsg.steve.repository.impl.WebUserRepositoryImpl;
 import de.rwth.idsg.steve.web.api.ApiControllerAdvice;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.context.annotation.Bean;
@@ -34,13 +35,10 @@ import org.springframework.security.config.annotation.web.configuration.EnableWe
 import org.springframework.security.config.http.SessionCreationPolicy;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
-import org.springframework.security.core.userdetails.User;
-import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.crypto.factory.PasswordEncoderFactories;
 import org.springframework.security.crypto.password.DelegatingPasswordEncoder;
 import org.springframework.security.crypto.password.PasswordEncoder;
-import org.springframework.security.provisioning.InMemoryUserDetailsManager;
 import org.springframework.security.web.AuthenticationEntryPoint;
 import org.springframework.security.web.SecurityFilterChain;
 import org.springframework.security.web.authentication.preauth.AbstractPreAuthenticatedProcessingFilter;
@@ -75,14 +73,8 @@ public class SecurityConfiguration {
     }
 
     @Bean
-    public UserDetailsService userDetailsService() {
-        UserDetails webPageUser = User.builder()
-                .username(CONFIG.getAuth().getUserName())
-                .password(CONFIG.getAuth().getEncodedPassword())
-                .roles("ADMIN")
-                .build();
-
-        return new InMemoryUserDetailsManager(webPageUser);
+    public UserDetailsService userDetailsService(WebUserRepositoryImpl webUserRepository) {
+        return webUserRepository;
     }
 
     @Bean
--- a/src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
+++ b/src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
@@ -20,8 +20,6 @@ package de.rwth.idsg.steve.config;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
-import de.rwth.idsg.steve.SteveConfiguration;
-import de.rwth.idsg.steve.repository.WebUserRepository;
 import de.rwth.idsg.steve.web.api.ApiControllerAdvice;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
@@ -37,7 +35,6 @@ import org.springframework.security.config.annotation.web.configuration.EnableWe
 import org.springframework.security.config.http.SessionCreationPolicy;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
-import org.springframework.security.core.userdetails.User;
 import org.springframework.security.crypto.factory.PasswordEncoderFactories;
 import org.springframework.security.crypto.password.DelegatingPasswordEncoder;
 import org.springframework.security.crypto.password.PasswordEncoder;
@@ -45,7 +42,6 @@ import org.springframework.security.web.AuthenticationEntryPoint;
 import org.springframework.security.web.SecurityFilterChain;
 import org.springframework.security.web.authentication.preauth.AbstractPreAuthenticatedProcessingFilter;
 
-import jakarta.annotation.PostConstruct;
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
@@ -63,24 +59,6 @@ import static de.rwth.idsg.steve.SteveConfiguration.CONFIG;
 @Configuration
 @EnableWebSecurity
 public class SecurityConfiguration {
-
-    private final WebUserRepository webUserRepository;
-
-    @PostConstruct
-    public void postConstruct() {
-        if (webUserRepository.hasUserWithAuthority("ADMIN")) {
-            return;
-        }
-
-        var user = User
-            .withUsername(SteveConfiguration.CONFIG.getAuth().getUserName())
-            .password(SteveConfiguration.CONFIG.getAuth().getEncodedPassword())
-            .disabled(false)
-            .authorities("ADMIN")
-            .build();
-
-        webUserRepository.createUser(user);
-    }
 
     /**
      * Password encoding changed with spring-security 5.0.0. We either have to use a prefix before the password to
--- a/src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
+++ b/src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
@@ -21,6 +21,7 @@ package de.rwth.idsg.steve.config;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
 import de.rwth.idsg.steve.SteveConfiguration;
+import de.rwth.idsg.steve.repository.WebUserRepository;
 import de.rwth.idsg.steve.web.api.ApiControllerAdvice;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
@@ -40,7 +41,6 @@ import org.springframework.security.core.userdetails.User;
 import org.springframework.security.crypto.factory.PasswordEncoderFactories;
 import org.springframework.security.crypto.password.DelegatingPasswordEncoder;
 import org.springframework.security.crypto.password.PasswordEncoder;
-import org.springframework.security.provisioning.UserDetailsManager;
 import org.springframework.security.web.AuthenticationEntryPoint;
 import org.springframework.security.web.SecurityFilterChain;
 import org.springframework.security.web.authentication.preauth.AbstractPreAuthenticatedProcessingFilter;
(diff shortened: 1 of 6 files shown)


src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
```diff
@@ -61,11 +60,31 @@
...
 @Configuration
 @EnableWebSecurity
 @Conditional(SteveProdCondition.class)
 public class SecurityConfiguration {
 
+    private final UserDetailsManager userDetailsManager;
+
+    @PostConstruct
+    public void postConstruct() {
+        String userName = SteveConfiguration.CONFIG.getAuth().getUserName();
```
juherr, line 73: If not yet working, why not make it optional and use a default value?

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
```diff
@@ -61,11 +60,31 @@
...
+        if (userDetailsManager.userExists(userName)) {
+            return;
+        }
+
+        var user = User
+            .withUsername(userName)
+            .password(SteveConfiguration.CONFIG.getAuth().getEncodedPassword())
```
juherr, line 80: Why not make it optional and generate a default password which will be logged like MySQL is doing.

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
```diff
@@ -61,11 +60,31 @@
...
 @EnableWebSecurity
 @Conditional(SteveProdCondition.class)
 public class SecurityConfiguration {
 
+    private final UserDetailsManager userDetailsManager;
+
+    @PostConstruct
+    public void postConstruct() {
+        String userName = SteveConfiguration.CONFIG.getAuth().getUserName();
+        if (userDetailsManager.userExists(userName)) {
```
juherr, line 74: In fact, it can be skipped if an admin user already exists.

juherr (approved):
LGTM, it will be an awesome feature.

I've just shared some ideas.

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
goekay: "not working" was referring to my first commit from yesterday. current state is fully working.

it even includes the optional migration of user/pwd from properties file to database. optional, because it only creates this user if the user with the same name does not exist, and hence it will be executed only for the first ever execution of this new logic.

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
goekay: see https://github.com/steve-community/steve/pull/1539/files#r1712974597

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
goekay: why? we are migrating from a state with a known username and pwd from properties file into DB. i can recreate the same world in DB. it would then not break installations. the data will be moved to database.

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
fnkbsi: If the migrated user from the config is deleted (replaced by a other web_user), doesn't the code recreate the user from the config at a restart of Steve? 
I would prefer skip if any admin is present in the db. In that case if someone accidentally deleted all web_user admins a simple restart recreates the standard admin

goekay: okay, fair enough, done in https://github.com/steve-community/steve/pull/1539/commits/ffd7bca874db6306ad561936c92eda02d3fb3ab6

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
juherr: Sorry, it was just a global idea, not an issue for that current PR.

I recommend having a default value with "convention over configuration" in mind.
I don't remember if it already works like that or not. I think it is not.

juherr: Forget it: `getAuth().getUserName()` should always return something (and manage a default value if needed)

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
juherr: Same idea as the username: "convention over configuration".
But it is not relevant here.

src/main/java/de/rwth/idsg/steve/repository/impl/WebUserRepositoryImpl.java
```diff
@@ -0,0 +1,226 @@
...
+     */
+    private void validateUserDetails(UserDetails user) {
+        Assert.hasText(user.getUsername(), "Username may not be empty or null");
+        validateAuthorities(user.getAuthorities());
+    }
+
+    /**
+     * Lifted from {@link JdbcUserDetailsManager#validateAuthorities(Collection)}
+     */
+    private void validateAuthorities(Collection<? extends GrantedAuthority> authorities) {
```
juherr, line 208: It could be static

src/main/java/de/rwth/idsg/steve/repository/impl/WebUserRepositoryImpl.java
```diff
@@ -0,0 +1,226 @@
...
+            return JSON.jsonOrNull(str);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Lifted from {@link JdbcUserDetailsManager#validateUserDetails(UserDetails)}
+     */
+    private void validateUserDetails(UserDetails user) {
```
juherr, line 200: It could be static

juherr (approved):

src/main/java/de/rwth/idsg/steve/repository/impl/WebUserRepositoryImpl.java
goekay: done in https://github.com/steve-community/steve/pull/1539/commits/960fa23089f37069dd2765428802d4f6e97f6f15

src/main/java/de/rwth/idsg/steve/repository/impl/WebUserRepositoryImpl.java
goekay: done in https://github.com/steve-community/steve/pull/1539/commits/960fa23089f37069dd2765428802d4f6e97f6f15

goekay:
@fnkbsi how do you want to continue from here. we have your solution/branch and my solution/branch. 

would you be okay with me merging this first, and you could rebase your branch to add the missing and web-related stuff on top?

juherr (approved):

fnkbsi:
> @fnkbsi how do you want to continue from here. we have your solution/branch and my solution/branch.
> 
> would you be okay with me merging this first, and you could rebase your branch to add the missing and web-related stuff on top?

@goekay: Go ahead

fnkbsi:
One more idea to discuss. Do we want a optional link (foreign_key) between user and web_user?

juherr:
@fnkbsi I don't see value add to link user and webuser, do you?

fnkbsi:
> @fnkbsi I don't see value add to link user and webuser, do you?

Maybe in some businesses cases. If a user gets also an web_user, the user can be managed on one website. If the user will be deleted, the web_user could be automatically deleted too.

I just want to address this idea before the db is changed.

juherr:
In my understanding, Steve serves as a CPO layer that manages the stations and where the web users act as the operators. 

Currently, Steve is not a multi-tenant software, meaning all users have the same permissions across all stations. 

However, if we were to link specific users or tags to web users, it would open the door to the EMSP world, making the system more intricate. 
Please don't! 🙏

goekay:
@fnkbsi i think this is an interesting proposition. maybe we can do it in future, if we want to expand the capabilities of steve.

i added the token column as a preparation in https://github.com/steve-community/steve/pull/1539/commits/93c9cc07e34721d09f244d48948060e4ddc0d63d, because i think web users and api clients are within the same realm of "technical/operator user". one is human, the other is computer.

connecting a "web user" with a "user" would connect the technical world with the ocpp world (business). careful planning would be needed, if we want to. therefore, i would not want to rush it.

@juherr i know what this means. it would be a big undertaking. i am not categorically against it. it is just a matter of resources (i.e. time, active developers etc.).

juherr:
IMO, if you go that way, you should start by linking web users to stations 🤷🏼‍♂️

goekay:
FYI to the readers of this PR from the future: 

if you are using a version of steve with this PR included, your `auth.user` and `auth.password` settings from internal/external properties file will be migrated to database during the first ever start-up. afterwards, only database contents will be used/respected. meaning: if you change these two fields from properties file _afterwards_, they will not be used/respected.

i plan to remove these two fields from properties file in a later version.

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
```diff
@@ -58,10 +59,29 @@
...
  */
 @Slf4j
+@RequiredArgsConstructor
 @Configuration
 @EnableWebSecurity
 public class SecurityConfiguration {
 
+    private final WebUserRepository webUserRepository;
+
+    @PostConstruct
```
juherr, line 69: Not sure if post construct of a spring configuration is the best place for logic.

Instead, what do you think about:

```java
@RequiredArgsConstructor
@Component
public static class AutoCreateAdminAtStartup {
    private final WebUserRepository webUserRepository;

    @EventListener
    public void handleContextStart(ContextStartedEvent cse) {
        if (webUserRepository.hasUserWithAuthority("ADMIN")) {
            return;
        }

        var user = User
            .withUsername(SteveConfiguration.CONFIG.getAuth().getUserName())
            .password(SteveConfiguration.CONFIG.getAuth().getEncodedPassword())
            .disabled(false)
            .authorities("ADMIN")
            .build();

        webUserRepository.createUser(user);
    }
}
```

src/main/java/de/rwth/idsg/steve/repository/WebUserRepository.java
```diff
@@ -0,0 +1,30 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package de.rwth.idsg.steve.repository;
+
+import org.springframework.security.provisioning.UserDetailsManager;
+
+public interface WebUserRepository extends UserDetailsManager {
```
juherr, line 23: That's a bit weird to hide a high-level API like UserDetailsManager behind a low-level API like a repository.

I mean, from a conceptual point of view, it sounds logical that a UserDetailsManager uses a repository but it sounds weird that a repository "is a" UserDetailsManager.
It is like `JdbcUserDetailsManager` extending `JdbcTemplate` instead of using it.

WDYT?

goekay: i agree with you.

at first i was also confused when i looked at the interfaces UserDetailsService and UserDetailsManager, and their implementations, such as JdbcDaoImpl and JdbcUserDetailsManager. JdbcDaoImpl and JdbcUserDetailsManager are pure repository implementations.

then, i implemented the necessary functions in WebUserRepositoryImpl and realized that like 95% of the code is DB access only. then i stopped worrying.

i can add some additional ceremony by adding WebUserService that uses a WebUserRepository etc. but a  WebUserService would just forward the call in most cases.

juherr: I think you should. At least it will be straight forward when you read the implementation.

src/main/java/de/rwth/idsg/steve/repository/impl/WebUserRepositoryImpl.java
```diff
@@ -0,0 +1,226 @@
...
+            .execute();
+    }
+
+    @Override
+    public boolean hasUserWithAuthority(String authority) {
+        JSON authValue = JSON.json("\"" + authority + "\"");
+        return ctx.selectOne()
+            .from(WEB_USER)
+            .where(condition("json_contains({0}, {1})", WEB_USER.AUTHORITIES, authValue))
+            .fetchOptional()
```
fnkbsi, line 117: fetchOptional(): use this only if the query returns at most one record. 
So if there a 2 or more admins it crashes. 
--> .fetch().isNotEmpty()

goekay: good catch, thanks.

goekay: fixed in [0992d5c](https://github.com/steve-community/steve/pull/1539/commits/0992d5cdc3f13ead84e1d09157b8d1bfdd80bfde)

i opted for count, since we dont need all these rows with 1. count is a better fitting usage.

src/main/java/de/rwth/idsg/steve/repository/WebUserRepository.java
goekay: done in https://github.com/steve-community/steve/pull/1539/commits/0992d5cdc3f13ead84e1d09157b8d1bfdd80bfde

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
goekay: done in https://github.com/steve-community/steve/commit/0992d5cdc3f13ead84e1d09157b8d1bfdd80bfde

however, i did it as 
```
@EventListener
public void afterStart(ContextRefreshedEvent event) {
   ...
}
```
since your suggested method was not being triggered. this is more in line with some other usage in the repo anyways.

juherr (approved):
LGTM 👍

PR #1544, opened 2024-08-16
Added website stuff for Web-User administration and additional request…

fnkbsi:
Adding websites, controller and some methodes for multi Web-Users


src/main/java/de/rwth/idsg/steve/repository/dto/WebUserOverview.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.Getter;
+
+@Getter
+@Builder
+public final class WebUserOverview {
+
+    private final Integer webUserPk;
+    private final Boolean enabled;
+    private final String webusername;
+    private final String[] autorithies;
```
goekay, line 30: typo in field name. and why `String[]`, which can be considered a low-level construct by modern practices, instead of `List<String>`

src/main/java/de/rwth/idsg/steve/repository/impl/WebUserRepositoryImpl.java
```diff
@@ -40,6 +48,7 @@
 public class WebUserRepositoryImpl implements WebUserRepository {
 
     private final DSLContext ctx;
+    private final ObjectMapper jacksonObjectMapper;
```
goekay, line 51: please do not leak ObjectMapper into repository layer.

my first impl was like this (as there was no WebUserService). but during the refactor, i moved all business decisions and non-database actions into service layer.

src/main/java/de/rwth/idsg/steve/web/dto/Statistics.java.bak
goekay, line 1: Statistics.java.bak ?

src/main/java/de/rwth/idsg/steve/web/controller/WebUsersController.java
```diff
@@ -0,0 +1,170 @@
...
+    public String addPost(@Valid @ModelAttribute("webuserForm") WebUserForm webuserForm,
+                          BindingResult result, Model model) {
+        if (result.hasErrors()) {
+            return "data-man/webuserAdd";
+        }
+
+        // password is Null, Blank/Empty or less than 8 Characters then don't add and show an Error
+        if (webuserForm.getPassword() == null) {
+            webuserForm.setPwerror(Boolean.TRUE);
+            return "data-man/webuserAdd";
```
goekay, line 101: this type of validation should be put into WebUserForm with validator annotations, and not into java code as plainly as you did. see other Form usages.

src/main/resources/webapp/WEB-INF/views/data-man/webuserAdd.jsp
```diff
@@ -0,0 +1,72 @@
...
+            <tbody>
+            <tr><td>Webusername:</td><td><form:input path="webusername" title="Web-Users Name"/></td></tr>
+            <tr><td>Password:</td><td><form:password path="password" title="Set the password"/></td></tr>
+            <tr><td>Password confirmation:</td><td><form:password  path="passwordComparison" title="Confirm the password"/></td></tr>
+            <tr><td>Roles:</td>
+                <td>
+                    <form><select id="myAuthoritiesList" name="authorities" path="authorities" title="List of roles/authoriies the web-user shall have.">
+                                    <option value="USER" >USER</option>
+                                    <option value="ADMIN" >ADMIN</option>
+                                    <option value="USER,ADMIN" >USER, ADMIN</option>
```
goekay, line 54: you are deciding and hardcoding business values in view layer. maybe this should be an java enum?

src/main/resources/webapp/WEB-INF/views/data-man/webuserDetails.jsp
```diff
@@ -0,0 +1,75 @@
...
+                <tr><td>Username:</td><td>${webuserForm.webusername}<form:hidden path="webusername" readonly="true"/></td></tr>
+                <tr><td>Password:</td><td><form:password path="password" title="If needed a new password can be set."/></td></tr>
+                <tr><td>Password confirmation:</td><td><form:password path="passwordComparison" title="Confirm the password, then push the update button" /></td></tr>
+                <tr><td>Roles:</td>
+                    <td>
+                        <select id="myRoleList" name="authorities" path="authorities" title="List of roles/authoriies the web-user has.">
+                            <option value="${webuserForm.authorities}" >${webuserForm.authorities}</option>
+                            <option value="USER" >USER</option>
+                            <option value="ADMIN" >ADMIN</option>
+                            <option value="USER,ADMIN" >USER, ADMIN</option>
```
goekay, line 50: see https://github.com/steve-community/steve/pull/1544/files#r1720550870

src/main/java/de/rwth/idsg/steve/repository/dto/WebUserOverview.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.Builder;
+import lombok.Getter;
+
+@Getter
+@Builder
+public final class WebUserOverview {
+
+    private final Integer webUserPk;
+    private final Boolean enabled;
+    private final String webusername;
```
goekay, line 29: pls use camel case: webusername -> webUsername

src/main/java/de/rwth/idsg/steve/service/WebUserService.java
```diff
@@ -153,6 +162,53 @@ public boolean hasUserWithAuthority(String authority) {
...
+
+    public void update(WebUserForm form) {
+        updateUser(toUserDetails(form));
+    }
+
+    public List<WebUserOverview> getOverview(WebUserQueryForm form) {
+        return webUserRepository.getOverview(form);
+    }
+
+    public WebUserForm getDetails(Integer webuserpk) {
```
goekay, line 179: to be consistent with other usages, pls use camel case: webuserpk -> webUserPk

src/main/java/de/rwth/idsg/steve/web/controller/WebUsersController.java
```diff
@@ -0,0 +1,170 @@
...
+
+    // -------------------------------------------------------------------------
+    // Paths
+    // -------------------------------------------------------------------------
+
+    private static final String QUERY_PATH = "/query";
+
+    private static final String DETAILS_PATH = "/details/{webuserpk}";
+    private static final String DELETE_ALL_PATH = "/delete/{webuserpk}";
+    private static final String UPDATE_PATH = "/update";
```
goekay, line 53: to be consistent with other usages, pls use camel case: webuserpk -> webUserPk

src/main/java/de/rwth/idsg/steve/web/dto/WebUserForm.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private String password;
+
+    private String passwordComparison;
+
+    private String authorities;
+
+    private Boolean pwerror;
+
+    private String apitoken;
```
goekay, line 48: - pls use camel case in fields
- pwerror should not be needed after refactoring to use validation annotations.

src/main/java/de/rwth/idsg/steve/web/dto/WebUserQueryForm.java
```diff
@@ -0,0 +1,58 @@
...
+@Getter
+@Setter
+public class WebUserQueryForm {
+
+    private Boolean enabled;
+
+    // Free text input
+    private String webusername;
+    private String roles;
+    private String apitoken;
```
goekay, line 40: pls use camel case in fields

src/main/java/de/rwth/idsg/steve/repository/impl/WebUserRepositoryImpl.java
```diff
@@ -53,12 +62,22 @@ public void createUser(WebUserRecord user) {
 
     @Override
     public void updateUser(WebUserRecord user) {
-        ctx.update(WEB_USER)
-            .set(WEB_USER.PASSWORD, user.getPassword())
-            .set(WEB_USER.ENABLED, user.getEnabled())
-            .set(WEB_USER.AUTHORITIES, user.getAuthorities())
-            .where(WEB_USER.USERNAME.eq(user.getUsername()))
-            .execute();
+        // There is not alway the need to change the password
```
juherr, line 65: Maybe password should have its own route, process and ui for modification.

src/main/java/de/rwth/idsg/steve/service/WebUserService.java
```diff
@@ -153,6 +162,53 @@ public boolean hasUserWithAuthority(String authority) {
...
+
+        if (ur == null) {
+            throw new SteveException("There is no user with id '%d'", webuserpk);
+        }
+
+        WebUserForm form = new WebUserForm();
+
+        form.setEnabled(ur.getEnabled());
+        form.setWebusername(ur.getUsername());
+        form.setPassword(ur.getPassword());
```
juherr, line 190: Not sure that password and token should be exposed to ui.

src/main/resources/webapp/WEB-INF/views/data-man/webuserAdd.jsp
juherr: Admin is supposed to have more rights than user. What the purpose to have many roles for one user?

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
```diff
@@ -86,6 +86,27 @@ public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Excepti
...
+                    .requestMatchers(prefix + "/webusers").hasAnyAuthority("USER", "ADMIN")
+                    .requestMatchers(prefix + "/webusers" + "/details/**").hasAnyAuthority("USER", "ADMIN")
+                    // users
+                    .requestMatchers(prefix + "/users").hasAnyAuthority("USER", "ADMIN")
+                    .requestMatchers(prefix + "/users" + "/details/**").hasAnyAuthority("USER", "ADMIN")
+                     //ocppTags
+                    .requestMatchers(prefix + "/ocppTags").hasAnyAuthority("USER", "ADMIN")
+                    .requestMatchers(prefix + "/ocppTags" + "/details/**").hasAnyAuthority("USER", "ADMIN")
+                     // chargepoints
+                    .requestMatchers(prefix + "/chargepoints").hasAnyAuthority("USER", "ADMIN")
```
juherr, line 100: As user can't create, update ou delete charpoints (and some others), the UI should be updated (at least remove delete and update button)

src/main/java/de/rwth/idsg/steve/repository/dto/WebUserOverview.java
fnkbsi: ... because of <private String[] fromJson(JSON jsonArray) >

src/main/resources/webapp/WEB-INF/views/data-man/webuserAdd.jsp
fnkbsi: > Admin is supposed to have more rights than user. What the purpose to have many roles for one user?

In general yes, but in some cases the user has privileges a admin does not. To keep this freedom for the future, I added the option that one web-user can have both roles.

src/main/java/de/rwth/idsg/steve/web/controller/WebUsersController.java
fnkbsi: Your suggestion collides at the moment with the update path, were we may or may not update the password.

src/main/java/de/rwth/idsg/steve/repository/dto/WebUserOverview.java
goekay: you could always convert this array into a list yourself though. here is a helper method that you can put into WebUserService:

```
    private List<String> fromJsonToList(JSON jsonArray) {
        return Arrays.stream(fromJson(jsonArray)).toList();
    }
```

i did the method with String[] because this was the simplest way to do for my use case.

src/main/resources/webapp/WEB-INF/views/data-man/webuserAdd.jsp
goekay: > in some cases the user has privileges a admin does not.

can you give examples?

i think the common practice is that an admin is a superset of user.

goekay:
hey @fnkbsi i took the liberty of doing some small code improvements in https://github.com/steve-community/steve/pull/1544/commits/d00c287fa2e5aabfb29356e7ad30435ea83faf80 and hope you are okay with it 🙏

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
goekay: > As user can't create, update ou delete charpoints (and some others)

i will play devil's advocate (as setup for my next argument): why not?

your argument has a very specific understanding/model of what a user and admin should do or do not. i find it really hard to make decisions in steve about permissions and roles matrix because each customer/CPO/company would have different requirements depending on their organisation structure.

i have no solution for this. i dont think we can make some decisions (in this generic version of steve) that will make people happy.

goekay: the perfectionist in me says.. we should make the permissions and roles configurable in web UI or something, but this would be a big undertaking.

src/main/resources/webapp/WEB-INF/views/data-man/webuserAdd.jsp
fnkbsi: hypothetical: the USER authority could be to manage the users (read, write of ocpp_tag, user) and the ADMIN authority manages only the charge-boxes. In this scenario there are some web-users which are allowed to do both.

juherr: Based on what I understand your goal, I would like to propose implementing an additional user profile: CPO_USER (chargebox), EMP_USER (tag/user), and ADMIN (with full rights and privileges).

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
juherr: Sure, why not?

Let me share my own experience where people (administrative persons but technical operators too) unintentionally break things because they can. 

But what if we could give them a way to only "read" data and help drivers?

At the same time, "users" could be drivers in different contexts. 

In this vision, users become helpful companions for drivers/them, providing information and simple commands that won't cause any damage. Meanwhile, admins will be in charge of managing the network of stations.

Now, the question arises: how would we differentiate between admins and users? Can you imagine any noticeable differences in the way they are currently implemented?

To the perfectionist: I think it is recommended that implementation of peripheral functionalities, such as the one in question or even IAM, be delegated to specialized software stacks. This ensures that the primary objective of the software solution remains the central focus, while benefiting from the seamless integration of a well-established stack that effectively addresses the requirement.
However, it is important to note that implementing this way will require additional effort and may introduce complexity to the existing system. This could potentially impact the ops part.

src/main/java/de/rwth/idsg/steve/web/controller/WebUsersController.java
juherr: If you remove it from the form, it should fix most of the issues.
The password could be provided as result of the user creation.

src/main/java/de/rwth/idsg/steve/repository/dto/WebUserOverview.java
juherr: I'm a bit old fashion but for array-to-list conversion, I recommend `Arrays.asList()` instead.

goekay: my refactor commit eliminated the root cause and tackled this from a different perspective. therefore, the need for array-to-list is gone.

src/main/resources/webapp/WEB-INF/views/data-man/webuserAdd.jsp
goekay: see https://github.com/steve-community/steve/pull/1544#discussion_r1720797188

because of this, i think we should not try to find a golden solution. only 2 authorities (USER and ADMIN) will not be enough. the permission (what) distribution between roles (who) will be different for each org.

juherr: If the division of read and write permissions between user and admin roles is not widely used, a straightforward approach would be to remove role-based access control in the first implementation and grant all users access to all privileges.

WDYT?

fnkbsi: I added a java enum, if some one don't want a combination of USER&ADMIN just out-comment it in the enum. But it work's if needed.

src/main/resources/webapp/WEB-INF/views/data-man/webuserDetails.jsp
fnkbsi: using enum

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
fnkbsi: Actually the button are still there, but USER landing on the NoAccess site. If someone with more *jsp experience can hide them, I would welcome this!

juherr: I can do some ui improvments but @geokay must define first what is expected for USER and ADMIN.

fnkbsi:
@goekay @juherr 
I try to login the site as 'USER' and get an Error 500 >>java.lang.StackOverflowError<< 
Login as 'ADMIN' works fine. 
Q1: Can you confirm this behavior?
Q2: Any hints whats causing the error and how we can fix it?

src/main/java/de/rwth/idsg/steve/config/SecurityConfiguration.java
juherr: Another option can be ADMIN, SUPERVISOR, USER.
With 
USER = readonly, except users,
SUPERVISOR = read+write, except users,
ADMIN = read+write all

Wdyt @goekay ? Please advice.

goekay: i wrote this in an [earlier comment](https://github.com/steve-community/steve/pull/1544#discussion_r1720797188), and i still think that way: probably, we should not make any decisions about this. 

we can create the basic framework (i.e. the preparation for decisions) and ability to create roles. what these roles can do and cannot do should be left to each user of steve.

there are so many aspects and functionalities, i.e. master data management (read and write) and various ocpp operations. i find it hard to come up with one matrix that will be _the_ solution.

otherwise, we will get questions/issues like: why cant my USER role do X?

fnkbsi: In the current state of the code creating roles is possible by adapting the enum. 
Changing the privileges can be done by adapting the SecurityFilterChain, which of course could be tricky (https://github.com/steve-community/steve/pull/1544#issuecomment-2314552615). 
Are comments in the source code sufficient? 

So for the basic framework solution configuration I suggest:
User --> allowed to see all website/data
Admin --> allowed to see all websites and is allowed to change the data
[User, Admin] is commented out in the enum. I would like to keep it as a hint that multiple roles are technically possible

src/main/java/de/rwth/idsg/steve/web/dto/WebUserForm.java
```diff
@@ -0,0 +1,62 @@
...
+    private String password = "";
+
+    @NotNull(message = "Password repetition is required")
+    @Size(min = 8, message = "The repeated password also requires 8 or more characters")
+    private String passwordComparison;
+
+    @AssertFalse(message = "The repeated password did not match!")
+    private Boolean pwError;
+
+    private String apiPassword = "";
```
fnkbsi, line 45: @goekay:  The API password is missing validation rules.  Would the rule "@Size(min=8) be enough?

karelcoudijzer-alberic:
Is there any timeline on when this PR will be released?

Issue #1540, opened 2024-08-12
Store Web API key in database


goekay:
#### Checklist

- [x] I checked other issues already and found no answer
- [x] I checked the documentation and found no answer
- [x] I am running the latest version and the feature i am requesting is not implemented there

#### Describe the problem you are trying to solve

currently, we are storing [web api key in properties file](https://github.com/steve-community/steve/blob/master/src/main/resources/config/prod/main.properties#L23-L24). this is problematic for multiple reasons: 
- it is not dynamic/flexible/easy-to-change since hardcoded as property
- it is a singleton value valid for all steve API access. it is the same for all types API users. a user _A_ and another user _B_ should not share the key
- we should be able to revoke/change individual API keys on a per-use basis

#### Describe the solution you'd like

since we are moving into database for multi-users and rbac (see #991, #1165 and #1539), i am thinking about adding another column `api_token` to `web_user`. this way, each user will have the possibility to access web UI and API. with this approach, we will start associating tokens with web users.

#### Describe alternative solutions or features you've considered 

- add two new columns (header key and value) to settings table in order o preserve current behaviour
- add a parallel `web_token` (or similar) table to have a separate modeling/process/flow to deal with tokens.

#### Additional context

...

goekay:
during the impl of https://github.com/steve-community/steve/pull/1545 i realized the following. 

- we have to switch from api key to basic auth: we were setting headers for api key like `STEVE-API-KEY: apiKey` so far. i can extract `apiKey` into database, but which user should i associate this with? so far, it was an application-global setting but not anymore. we could just search for `apiKey` within all users... and authenticate this call whenever we find a match. this would probably be a security issue. therefore, these keys need to be specific for users. which nicely fits the basic auth mechanism, and if we do the switch, the key is not a key anymore but rather a password.
- `api_token` in database actually represents an "api password". this is interesting, and allows an admin/manager to set different passwords for UI and API. both can be set to the same value. i see this as an added expressiveness. is this an overkill?

juherr:
It is logical to consider using Basic Authentication when treating the `api_token` as an API password.
However, as an API integrator, I have not encountered such a method of API authentication before. 

It would be beneficial to explore dedicated stacks specifically designed for this purpose, such as Keycloak and other similar options.

In my view, offering a default implementation that is quick and rudimentary can be helpful, as long as it doesn't limit the option of replacing it with a more secure and improved solution in the future.

goekay:
[...]

PR #1545, opened 2024-08-17
switch to basic auth for API access

goekay:


src/main/java/de/rwth/idsg/steve/service/WebUserService.java
```diff
@@ -77,6 +82,10 @@ public void afterStart(ContextRefreshedEvent event) {
         this.createUser(user);
     }
 
+    private void moveApiTokenFromConfigToDatabase() {
+        // TODO
```
goekay, line 86: will implement later

src/main/java/de/rwth/idsg/steve/config/ApiAuthenticationManager.java
```diff
@@ -0,0 +1,116 @@
...
+    @Override
+    public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+        String username = (String) authentication.getPrincipal();
+        String apiPassword = (String) authentication.getCredentials();
+
+        if (Strings.isNullOrEmpty(username) || Strings.isNullOrEmpty(apiPassword)) {
+            throw new BadCredentialsException("Required parameters missing");
+        }
+
+        UserDetails userDetails = webUserService.loadUserByUsernameForApi(username);
```
goekay, line 68: this impl is a bit inefficient right now, since we go to database for each API call to validate the credentials

src/main/java/de/rwth/idsg/steve/config/ApiAuthenticationManager.java
```diff
@@ -0,0 +1,116 @@
...
+            authException.getMessage()
+        );
+
+        response.setStatus(status.value());
+        response.setContentType(MediaType.APPLICATION_JSON_VALUE);
+        response.getWriter().print(jacksonObjectMapper.writeValueAsString(apiResponse));
+    }
+
+
+    private boolean areValuesSet(UserDetails userDetails) {
```
juherr, line 103: Could be static

juherr (approved):
LGTM from a technical pov.
But why do you prefer basic auth instead of dedicated header or bearer token for api auth?

goekay:
> But why do you prefer basic auth instead of dedicated header or bearer token for api auth?

pls see https://github.com/steve-community/steve/issues/1540#issuecomment-2294698082

src/main/java/de/rwth/idsg/steve/config/ApiAuthenticationManager.java
goekay: added cache in https://github.com/steve-community/steve/pull/1545/commits/e646b474007cdfe9c0f5b931bf7718500a8f4373 with a TTL of 10 min

src/main/java/de/rwth/idsg/steve/config/ApiAuthenticationManager.java
```diff
@@ -99,8 +113,19 @@ public void commence(HttpServletRequest request,
         response.getWriter().print(jacksonObjectMapper.writeValueAsString(apiResponse));
     }
 
+    private UserDetails getFromCacheOrDatabase(String username) {
```
juherr, line 116: Why do you set the cache logic here instead of `webUserService.loadUserByUsernameForApi`?

I don't know if it is a choice to not use it but Spring has its own way of doing caches: https://spring.io/guides/gs/caching 
And you can still use guava under to wood if you want it: https://docs.spring.io/spring-framework/docs/4.2.x/javadoc-api/org/springframework/cache/guava/GuavaCacheManager.html

src/main/java/de/rwth/idsg/steve/service/WebUserService.java
goekay: my WIP approach/thinking:

check whether some web_user has api_password
- yes: no need to insert anything. do not change existing setups.
- no: assign the api_token as api_password to all web_users with ADMIN authority. (since current default setup has only one user, and that is admin, we will set associate this api_password with it) 

wdyt?

juherr: I am just thinking. Do you need to migrate the previous token? The change is already breaking with the auth mode change and users will have to adapt the integration. So the effort to create an API password won't be so hard.

goekay: sure, i would have no problem with this.

src/main/java/de/rwth/idsg/steve/config/ApiAuthenticationManager.java
juherr: @goekay Just for my technical knowledge because I didn't use them before: why do you use guava directly and not hidden behind spring cache?

juherr: Thanks for the answer! 👍 

> why introduce spring magic?

That could be true for almost every spring import ;)

goekay:
@juherr if you have no objections or no more comments, i want to merge this.

juherr (approved):
LGTM

faculoyarte:
Sorry guys, but I'm having trouble figuring out how to create an api_password on a user or set up admin users in the SteVe web UI. I had the API working previously using webapi.key = STEVE-API-KEY and webapi.value, but now I'm getting a 401 error. Could someone guide me on how to resolve this?"

goekay:
hey @faculoyarte, the user you are showing on the screenshot is the _end user_, i.e. the customer that has an EV and RFID card and wants to use the stations.

the user we added in this PR is the _web user_, i.e. the operations person that manages stations, someone that belongs to a CPO maybe. this is the person that has access to steve's web ui to do things. the web user gets an api_password with this PR.

therefore, these two things are disconnected. there is [another PR](https://github.com/steve-community/steve/pull/1544) that will make it available to update/change properties of a web user. therefore, currently the only way to do is to directly modify database tables.

faculoyarte:
Perfect, thanks. @goekay

PR #1556, opened 2024-08-24
refactor DataSource

goekay:
* make DataSource a spring bean
* move checkJavaAndMySQLOffsets() into GenericRepository
* necessary consequential changes

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/config/BeanConfiguration.java
+++ b/src/main/java/de/rwth/idsg/steve/config/BeanConfiguration.java
@@ -29,7 +29,6 @@ import de.rwth.idsg.steve.SteveConfiguration;
 import de.rwth.idsg.steve.service.DummyReleaseCheckService;
 import de.rwth.idsg.steve.service.GithubReleaseCheckService;
 import de.rwth.idsg.steve.service.ReleaseCheckService;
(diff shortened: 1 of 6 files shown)


src/main/java/de/rwth/idsg/steve/config/BeanConfiguration.java
```diff
@@ -126,9 +125,7 @@ private void initDataSource() {
      * - http://stackoverflow.com/questions/32848865/jooq-dslcontext-correct-autowiring-with-spring
      */
     @Bean
-    public DSLContext dslContext() {
-        initDataSource();
-
+    public DSLContext dslContext(DataSource dataSource) {
```
juherr, line 128: You can avoid the parameter if you get bean by calling `dataSource()` like in many spring samples.

src/main/java/de/rwth/idsg/steve/repository/impl/GenericRepositoryImpl.java
```diff
@@ -52,6 +59,25 @@ public class GenericRepositoryImpl implements GenericRepository {
 
     @Autowired private DSLContext ctx;
 
+    @EventListener
```
juherr, line 62: It is mixing repository and service layer.

Why not move the logic into a service and use a `getMySQLOffsets()` provided by the repository?

Issue #100, opened 2018-12-19
OCPP 1.6-J Security


V2G-UK:
The OCA have just back ported OCPP 2.0 features to OCPP 1.6-J only. Do the SteVe team have any plans to support this new part of the standard? See:

http://www.v2g-evse.com/2018/12/18/open-charge-alliance-enhances-ocpp-1-6-security/

> Secure communication and operation is a critical aspect of Electric Vehicle Charging Infrastructure.
> 
> To further assist the industry the Open Charge Alliance now publishes a white paper to describe a standard way to address security using OCPP 1.6-J.

According to the new "white paper":

> This document is for OCPP 1.6-J (JSON over WebSockets) only, OCPP-S (SOAP) is NOT supported. This document was started, as it is seen as a simple stap to port OCPP 2.0 security to OCPP 1.6. But as OCPP 2.0 only support JSON over WebSockets (not SOAP), this document is also written for OCPP 1.6-J only. Adding SOAP to this document would have taken a lot of work and review by security experts.
> 
> This document is based on OCPP 2.0. To help developers that are implementing both 1.6J security improvement and OCPP 2.0, we have kept the Use Case numbering from OCPP 2.0. So when implementing for example Use Case N01, it is the same use case in this document as in the 2.0 specification.

lategoodbye:
Since this feature is requested regularly, maybe we should discuss some specific points here:

- What is the expected security profile?
- Is it possible to implement a conform Basic Auth in Steve (reject invalid credential before upgrading the websocket)? 
- Which messages must be implemented (e.g. handle SecurityEventNotification from the ChargePoint)?

andle:
@goekay Thanks for the great work as always! 

Just wanted to (poke) see if there were any updates on this?
I'm finding more CPO's requesting Basic Auth and wanted to be able to test my EVSE's implementation.

I like your idea about adding username/password configuration into the ChargePoint's page (can't find which issue discussed that).

nikhilkandari77:
> The OCA have just back ported OCPP 2.0 features to OCPP 1.6-J only. Do the SteVe team have any plans to support this new part of the standard? See:
> 
> http://www.v2g-evse.com/2018/12/18/open-charge-alliance-enhances-ocpp-1-6-security/
> 
> > Secure communication and operation is a critical aspect of Electric Vehicle Charging Infrastructure.
[...]

PR #1574, opened 2024-09-11
Warn about security risks

juherr:
Related to #100


goekay:
[there is a nuance though: ](https://github.com/steve-community/steve/issues/100#issuecomment-448998151)

> in the mean time, you can use TLS with steve for communication with your SOAP or JSON stations already. just install the necessary certificates in your java keystore and in your charging stations. then, you can use the path prefixes wss:// for JSON and https:// for SOAP stations. however, a certificate management as described in the new spec is not present.

PR #1577, opened 2024-09-15
add APIs link to navigation header

goekay:


src/main/resources/webapp/WEB-INF/views/00-header.jsp
```diff
@@ -73,6 +73,7 @@
                     <li><a href="${ctxPath}/manager/settings">SETTINGS</a></li>
                     <li><a href="${ctxPath}/manager/log">LOG</a></li>
                     <li><a href="${ctxPath}/manager/about">ABOUT</a></li>
+                    <li><a href="${ctxPath}/manager/swagger-ui/index.html">APIs</a></li>
```
goekay, line 76: all other entires are all uppercase. here, "APIS" looks weird, and hence this "APIs". but it breaks uniformity.

just "API" might be an option, as well.

juherr (approved):

PR #1682, opened 2025-01-26
Improve query functionality

goekay:
1st commit adds a new functionality in terms of a new field to query by.

2nd commit addresses an annoyance and confusion when using swagger. apparently, another solution is setting the property `springdoc.use-fqn=true`. but then the class name uses the complete package name, which can be considered an unnecessary leak of impl detail.


juherr:
@goekay Fyi about the 2nd commit, using [`@Schema(name="MyDto")`](https://docs.swagger.io/swagger-core/v2.0.0-RC3/apidocs/io/swagger/v3/oas/annotations/media/Schema.html#name--) on the class should work too.

goekay:
thanks, i stumbled upon that as well, but found it less robust.

Issue #1721, opened 2025-03-22
Swagger UI for the APIs unreachable


goekay:
#### Checklist

- [x] I checked other issues already, but found no answer/solution
- [x] I checked the documentation and wiki, but found no answer/solution
- [x] I am running the latest version and the issue still occurs
- [x] I am sure that this issue is about SteVe (and not about the charging station software or something unrelated to SteVe)

#### Specifications
```
SteVe Version     : ...
Operating system  : ...
JDK               : ...
Database          : ...
```

#### Expected Behavior

Swagger UI for the APIs is always reachable.

#### Actual Behavior

Swagger UI for the APIs become unreachable after some time of the deployment.

<img width="754" alt="Image" src="https://github.com/user-attachments/assets/91dc24f2-1af0-4cfb-9890-051deb6eba42" />

#### Additional context

I assume this to be connected with the path pointing at a file `.../index.html` (which can be interpreted as a static resource). Maybe using a path without file pointers could be the solution

goekay:
i kinda assume/hope that this issue will be gone after https://github.com/steve-community/steve/pull/1822

PR #1730, opened 2025-03-25
refactor

goekay:
remove some low-value code ceremony and simplify


juherr (approved):

goekay:
context and my behind-the-scenes thinking: we might be migrating to spring boot, after issues like [this](https://github.com/steve-community/steve/issues/1721).

juherr:
@goekay I believe using Spring Boot would be a smart move, especially since there’s already a well-executed fork that implemented it successfully.

goekay:
i believe you are referring to [this discussion](https://github.com/steve-community/steve/discussions/1606). i would welcome a PR but here was none :)

goekay:
here we go https://github.com/steve-community/steve/issues/1733 

:)

Issue #1734, opened 2025-03-28
Migrate to Spring Boot - Switch to Logback


goekay:
I am not sure how useful Log4j2 is, and it comes with a lot of pom.xml ceremony + com.lmax.disruptor as library.

Considering that Logback is default for Spring Boot, we should switch.

PR #1736, opened 2025-03-29
switch from log4j2 to logback (#1734)

goekay:
FYI @juherr 

notes: tried to keep most logging behaviour as it was
- same logging pattern/format
- same log level configs
- same directory & rolling file config with the exception of some rolling file policy aspects:
  - i think `SizeBasedTriggeringPolicy size="10 MB"` was unnecessary
  - max number of files to keep was set to 20. this incorporates the days and multiple files per day (due to 10MB size limit). now, we keep max history of 30 days with a total size cap of 3GB.


pom.xml
```diff
@@ -497,27 +498,13 @@
...
             <dependency>
                 <groupId>org.junit</groupId>
                 <artifactId>junit-bom</artifactId>
                 <version>5.12.1</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
-            <dependency>
-                <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-bom</artifactId>
```
juherr, line 516: Using bom allows you to set the version only once.

juherr (approved):
👍

pom.xml
goekay: i know, but it also introduces some indirection (i.e. _"pff... no version here, where is the version coming from, gaddamit"_). since, all logging stuff will come shortly from `spring-boot-starter-logging` i opted for simplicity and removing code.

PR #1744, opened 2025-04-01
Refactor application profiles to follow the Spring Boot convention

### LINKED ISSUE #1735: Migrate to Spring Boot - Refactor application profiles
Currently, we have https://github.com/steve-community/steve/tree/master/src/main/resources/config with different folders as profiles and some (once again) some pom.xml ceremony to select the profile and resource filtering.

Considering Spring Boot expects multiple profile-related files to co-exist with different name postfixes (such as application-dev.properties and application-prod.properties etc.), we should switch to this convention.

juherr:
Fix #1735


juherr:
Closing this PR due to inactivity (no response in 4 months). Feel free to reopen if needed.
See: https://github.com/juherr/evolve/pull/17

goekay:
hey, sorry for late response. do you still want feedback and be involved to merge this?

juherr:
I'm mostly interested in your take once you've dug into the code.
Skipping the config file at build time is easy.
The hard part is untangling the configuration from its usage — especially because it's wired through an enum singleton pattern.
If you see another clean path, I'd love to hear it.

goekay:
my general feedback:
- pls do not go into override territory. keep the properties file contents as they are. reason: we will also migrate to YML after spring boot. then, overriding will come out of the box from spring boot.
- it looks like many logging configs are the same (for ex: docker and kubernetes). since you are decoupling spring profile from the actual file (in logback-spring), we can keep X profiles while some of them referencing the same logback config. in short: a consolidation of logback files is possible.
- filtering of resources in pom.xml is getting complicated. i would suggest to just include everything under resources during build.

juherr:
You’re right. In my latest rework, I kept only one configuration:  
https://github.com/juherr/evolve/tree/main/steve/src/main/resources/config  

I handle the differences via Maven profiles:  
https://github.com/juherr/evolve/blob/main/steve/pom.xml#L265-L291  

As for Docker/Kubernetes, I believe they should match the production environment.  
The Dockerfile has been polished to support exactly that:  
https://github.com/juherr/evolve/blob/main/Dockerfile

goekay:
should i take over to apply the changes from my feedback?

juherr:
Yes, please go ahead and take care of it 👍

Issue #1733, opened 2025-03-28
Migrate to Spring Boot


goekay:
Solution approach: Prepare the codebase such that the delta between what Steve does and Spring Boot as default provides is minimal. Afterwards, the actual change should be as non-event as possible.

juherr:
What about the data layer? 
Move to Spring data or keep jooq?

goekay:
i don't aim to rewrite controller, service or repository layers. the goal is to use spring boot's easy bootstrapping instead of hand-wiring. plus, jooq has enough/good spring boot interop.

juherr:
@andortoth As you already did the job once, do you have any advice or feedback that can help?

juherr:
@goekay Could you review [this commit](https://github.com/parklapp/steve-pluggable/commit/4265bdee62569cc1eb67cab238043fa4b4aad8ac) and let me know what should be backported?
From what I understand, the main challenge in migrating to Spring Boot is the startup process involving configuration.

goekay:
this is such a heavy commit. need to do it later (since it needs attention/time).

> main challenge in migrating to Spring Boot is the startup process involving configuration.

yeah, but most of the config is coming from spring beans and `@Configuration`s anyways. this does not change when we have spring boot. the main challenge should be the following.

currently, whenever steve starts (Application's main method), it starts a plain Jetty server(!). within that we start and configure spring (in SteveAppContext). so, spring's context is only living in a small world.

spring likes to be the mothership though. it should be the first thing to start, and everything else should live within that. so we have to change
- Application -> Jetty -> Spring

to
- Application -> Spring -> Jetty (implicitly, by spring boot)

---

as a scratchpad, i managed to migrate steve to spring boot in a really dirty testing-the-waters way, and then threw it away. spring boot owned everything, jetty was there, DB via jooq as well. app was starting successfully.

so, as i said, as long as we don't touch the layers and change business logic (this heavy commit also does other things), this should not be that hard.

goekay:
to reduce complexity and keep the scope focused, i dont aim to touch the enum class `SteveConfiguration` and change it to a spring bean. not yet. not within this one.

andortoth:
> [@andortoth](https://github.com/andortoth) As you already did the job once, do you have any advice or feedback that can help?

our version was a major overhaul, because of some business requirements that might not suit the Steve library and its design philosophy, so we decided not to make a pull request of it

[...]

PR #1745, opened 2025-04-01
Move webapp folder in the default Spring Boot location

juherr:
Related to #1733


goekay:
could you please elaborate the changes?

juherr:
It is just moving the folder into the common folder used by spring boot for jsp.
I didn't see any side effect of the changes.

goekay:
> It is just moving the folder into the common folder used by spring boot for jsp.
I didn't see any side effect of the changes.

i just question this claim about common/default folder by spring boot, because i did encounter such a thing in my experience.

juherr:
As I understand it:
* When using JSP with Spring Boot, a WAR package is required, which in turn relies on the maven-war-plugin. (https://docs.spring.io/spring-boot/reference/web/servlet.html#web.servlet.embedded-container.jsp-limitations)
* The default `warSourceDirectory` for this plugin is `src/main/webapp`, and this value is not overridden by `spring-boot-starter-parent` (https://maven.apache.org/plugins/maven-war-plugin/war-mojo.html#warSourceDirectory)

juherr:
Closed in favor of migrating to Thymeleaf, which is a more appropriate next step.
See: https://github.com/juherr/evolve/pull/5

goekay:
hey @juherr could you please revert the deletion and reopen this PR?

juherr:
@goekay done. Feel free to update it.

goekay:
fixing conflicts was not fun. i abandoned it. instead, i applied the same changes on top of current HEAD in https://github.com/steve-community/steve/pull/1821

as always, thanks for the help @juherr and sorry for the unnecessary effort/trouble i caused.

PR #1817, opened 2025-09-16
Refactor constructor injection

goekay:
This PR is the spiritual successor of https://github.com/steve-community/steve/pull/1226 by @juherr. The original PR was closed and feature branch deleted unfortunately... because of my lack of timely feedback/reaction. I apologise for it.

That being said, this PR makes some different decisions in some occasions, where various approaches are technically valid and the end result depends on, more or less, style (i.e. the way to break the cyclic dependencies and simplification of classes during the AbstractWebSocketEndpoint refactor).

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/ocpp/ws/pipeline/OutgoingCallPipeline.java
+++ b/src/main/java/de/rwth/idsg/steve/ocpp/ws/pipeline/OutgoingCallPipeline.java
@@ -20,8 +20,6 @@ package de.rwth.idsg.steve.ocpp.ws.pipeline;
 
 import de.rwth.idsg.steve.ocpp.ws.FutureResponseContextStore;
 import de.rwth.idsg.steve.ocpp.ws.data.CommunicationContext;
-import lombok.extern.slf4j.Slf4j;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


juherr:
That's great to see it moving forward. Nice work!  
I'll take a quick look at the diff, but aside from style, could you explain the main differences compared to my version?  

On my side, I've been pushing the modularization further in [evolve](https://github.com/juherr/evolve), so feel free to ping me if I can help.

juherr:
FYI, I can restore any branch I’ve used in a PR, even if it was removed:  
image  

So just ping me if you’d like to revisit any of my previous PRs.

goekay:
> I'll take a quick look at the diff, but aside from style, could you explain the main differences compared to my version?

the two biggest differences are
- the way to break the cyclic dependencies
- simplification of classes during the AbstractWebSocketEndpoint refactor

because i chose a fundamentally different approach.

i did not mean style in terms of aesthetics actually. i meant the general design of the code. in many cases i took the path of least resistance trying to minimize the delta instead of doing some really aggressive changes. 

for example (from your PR):
```
@Getter(AccessLevel.PROTECTED)
private final OcppTagService ocppTagService;
```
having a private class member and encapsulating access via getter is too much _"by the book"_ IMO. i find it perfectly okay to have a `protected` class member such that child classes can access.

stuff like this.

juherr:
> instead of doing some really aggressive changes

Yes, I see what you mean — that’s exactly why I didn’t propose my latest changes.  
Still, I’d be glad if you could review them and challenge my approach.

goekay:
any feedback/objections @juherr ?

juherr:
I didn’t notice any major differences from the original PR. LGTM 👍

Issue #1819, opened 2025-09-17
Migrate to Spring Boot - Refactor as preparation


goekay:
Start refactoring classes and how we interact with Spring in general, such that the actual trigger step of using Spring-Boot becomes a small step. 

Criteria:
- Change how we wire Spring stuff
- No functional change: Current Spring app can be built and can run.

Inspiration: [This commit](https://github.com/steve-community/steve/commit/ccd34c6a0dad671158ef96881e5e9fa19be134e6) gave me the idea to continue such efforts.

PR #1820, opened 2025-09-18
Prep for Spring Boot migration

goekay:
https://github.com/steve-community/steve/issues/1819


pom.xml
```diff
@@ -518,6 +518,12 @@
             <groupId>org.springdoc</groupId>
             <artifactId>springdoc-openapi-starter-webmvc-ui</artifactId>
             <version>2.8.13</version>
+            <exclusions>
+                <exclusion>
+                    <artifactId>tomcat-embed-el</artifactId>
```
juherr, line 523: Something feels off here: you removed the EL implementation without adding an alternative.  
Without one, `OcppTagsRestControllerTest` will likely fail.  

Maybe `org.glassfish:jakarta.el` is still pulled in as a transitive dependency somewhere, but I haven’t been able to track down from where yet.  
Do you have any clue?

goekay: tests were not failing. otherwise, i would not have merged. i also have run the app and QA'ed the UI. it was working.

i removed it, because we had version conflicts. we had 2 versions of the same dependency. the other one was a transitive dependency as well.

juherr: Sure, you don’t see any test failures, but I do, since dependencies are managed differently across modules.  
Appreciate your confirmation that `tomcat-embed-el` is still the EL implementation.

goekay: should i remove it in my spring boot PR? i think exclusion was relevant for the prep (before spring boot). i can imagine it is redundant afterwards.

goekay: https://github.com/steve-community/steve/pull/1822/commits/0fd6b44d17fde8a98ff9c9d4a863aa1f7c29e832

juherr: As mentioned, `tomcat-embed-el` is a runtime dependency and could be swapped with another implementation.  
It’s clearer now, and the intent is well expressed. 👍

PR #1839, opened 2025-10-01
Feature/ocpp20

fmandache:
OCPP20


SvevoF:
Hi!

I'm currently working on an implementation of the OCPP 2.0.1 on an EVSE. I've seen that the `BootNotificationResponse` does not follow the OCPP2.0.1 JSON schema. More precisely:

```json
[3,"4798e4ef-2614-4858-a14f-8351e1727711",
{"currentTime":1704067200.0,
"interval":300,
"status":"Accepted",
"statusInfo": {"additionalInfo":
"OCPP 2.0.1 support is experimental",
"reasonCode":"OCPP20Experimental"}}]
```

reports an integer value for the `"currentTime"` key. The specification requests a `string`.
I've also seen that the tests do not check for the format, so that's maybe why it slipped through.

I've tried to change it myself but my knowledge of Java is very limited.
I've also tried to open an issue on the forked repo but it's not possible.

Thanks a lot in advance for your help, @fmandache !

Issue #1844, opened 2025-10-01
Regression: websocket connections close after 30 secs


goekay:
came after spring boot migration. apparently, [this configurer](https://github.com/steve-community/steve/blob/master/src/main/java/de/rwth/idsg/steve/config/WebSocketConfiguration.java#L83-L86) is not working.

```
[ERROR] 2025-10-01 19:20:26,814 de.rwth.idsg.steve.ocpp.ws.WebSocketLogger (Scheduler-1850322961-1) - [chargeBoxId=HELLO, sessionId=5c72ed47-fffe-3dd3-cee5-bb07a87fbacd] Transport error
org.eclipse.jetty.websocket.api.exceptions.WebSocketTimeoutException: Connection Idle Timeout
        at org.eclipse.jetty.websocket.common.JettyWebSocketFrameHandler.convertCause(JettyWebSocketFrameHandler.java:436)
        at org.eclipse.jetty.websocket.common.JettyWebSocketFrameHandler.onError(JettyWebSocketFrameHandler.java:240)
        at org.eclipse.jetty.websocket.core.WebSocketCoreSession.lambda$closeConnection$2(WebSocketCoreSession.java:262)
        at org.eclipse.jetty.server.handler.ContextHandler$ScopedContext.run(ContextHandler.java:1524)
        at org.eclipse.jetty.server.handler.ContextHandler$ScopedContext.run(ContextHandler.java:1511)
        at org.eclipse.jetty.websocket.core.server.internal.AbstractHandshaker$1.handle(AbstractHandshaker.java:179)
        at org.eclipse.jetty.websocket.core.WebSocketCoreSession.closeConnection(WebSocketCoreSession.java:262)
        at org.eclipse.jetty.websocket.core.WebSocketCoreSession.lambda$sendFrame$7(WebSocketCoreSession.java:503)
        at org.eclipse.jetty.util.Callback$3.succeeded(Callback.java:170)
        at org.eclipse.jetty.websocket.core.util.TransformingFlusher.notifyCallbackSuccess(TransformingFlusher.java:195)
        at org.eclipse.jetty.websocket.core.util.TransformingFlusher$Flusher.process(TransformingFlusher.java:152)
        at org.eclipse.jetty.util.IteratingCallback.processing(IteratingCallback.java:253)
        at org.eclipse.jetty.util.IteratingCallback.iterate(IteratingCallback.java:232)
[...]

Issue #1838, opened 2025-10-01
Regression: AbstractTypeStore is empty after spring boot migration


goekay:
https://github.com/steve-community/steve/discussions/1836

we use [this](https://github.com/steve-community/steve/blob/master/src/main/java/de/rwth/idsg/steve/ocpp/ws/AbstractTypeStore.java#L96-L99) to discover concrete classes to be used for the `requestClassMap`. it looks like this ClassLoader usage of guava does not play nice with spring boot: https://github.com/spring-projects/spring-boot/issues/5338

running the app plainly, in IDE, works fine. this is the reason for tests to pass i guess. but running the final war package is problematic.

from https://github.com/google/guava/issues/2152#issuecomment-441236717:

> The problem is that Guava doesn't understand JARs in JARs.

PR #1845, opened 2025-10-01
make DefaultHandshakeHandler a bean (#1844)

goekay:
if it is not a bean, DefaultHandshakeHandler.setServletContext is not called. DefaultHandshakeHandler delegates the call to RequestUpgradeStrategy.setServletContext. JettyRequestUpgradeStrategy's setServletContext calls the webSocketConfigurer.

this call was missing in the chain. we need it to be called since webSocketConfigurer contains config for extended idle timeout.

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/config/WebSocketConfiguration.java
+++ b/src/main/java/de/rwth/idsg/steve/config/WebSocketConfiguration.java
@@ -28,6 +28,7 @@ import de.rwth.idsg.steve.web.validation.ChargeBoxIdValidator;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.eclipse.jetty.websocket.core.WebSocketConstants;
+import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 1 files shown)


juherr:
@goekay Maybe we should move the tests using the real database from *unit* to *integration* tests.  
That would simply mean renaming them from `*Test` to `*IT` and starting Steve at the `pre-integration-test` phase.  

I don’t recall if the Failsafe plugin is already enabled by default or if we need to add it manually.  
What’s your take on this approach?

goekay:
I am neither for or against it. I don't mind. I find the separation just cosmetics/semantics in our case. We will still run all tests always.

But I cannot see how your suggestion is related to this issue/PR? how will it prevent this mistake in future?

juherr:
Sorry, my previous explanation wasn’t clear. This change will help prevent similar issues in the future when running the tests on a different JVM.

goekay:
is it possible that you actually [mean the other issue](https://github.com/steve-community/steve/issues/1838)?

Issue #1855, opened 2025-10-20
Change data columns in transactions table/page


goekay:
I want to add "User ID" to the overview and details pages, such that a person can connect the dots: who was the user of the Ocpp Id Tag that did the transaction?

But, I am concerned that the table will look crowded. Consequently, I want to remove the "Stop Reason" since I think it is a not-so-important and not-much-useful info at this overview level. It still exists, on details page anyways.

Changes:
- Add "User ID" to transactions overview page
- Add "User ID" to transaction details page
- Remove "Stop Reason" from transactions overview page.

juherr:
It makes sense on the page, but in the table it might lead to mixing CPO and EMSP data in the same database.  
I’d suggest keeping those contexts separated to avoid unnecessary complexity later on.

goekay:
table = html table on the page :)

juherr:
LGTM!

PR #1857, opened 2025-10-20
Change data columns in transactions table/page

goekay:
https://github.com/steve-community/steve/issues/1855

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/repository/impl/TransactionRepositoryImpl.java
+++ b/src/main/java/de/rwth/idsg/steve/repository/impl/TransactionRepositoryImpl.java
@@ -43,6 +43,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import static de.rwth.idsg.steve.repository.impl.RepositoryUtils.ocppTagByUserIdQuery;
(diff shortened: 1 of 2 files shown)


src/main/java/de/rwth/idsg/steve/repository/impl/TransactionRepositoryImpl.java
```diff
@@ -209,162 +268,76 @@ public TransactionDetails getDetails(int transactionPk) {
...
 
         if (form.isOcppIdTagSet()) {
-            selectQuery.addConditions(TRANSACTION.ID_TAG.eq(form.getOcppIdTag()));
+            conditions.add(TRANSACTION.ID_TAG.eq(form.getOcppIdTag()));
         }
 
         if (form.isUserIdSet()) {
-            var query = ocppTagByUserIdQuery(ctx, form.getUserId());
-            selectQuery.addConditions(TRANSACTION.ID_TAG.in(query));
+            conditions.add(USER_OCPP_TAG.USER_PK.eq(form.getUserId()));
```
goekay, line 297: since we are left-joining with the `USER_OCPP_TAG` table, we have direct access to `userId` in the original query now. `ocppTagByUserIdQuery(ctx, form.getUserId());` is not needed anymore.

Issue #1933, opened 2026-01-08
Improve the Docker-Compose


KjellWolf:
#### Checklist

- [ x ] I checked other issues already and found no answer
- [ x ] I checked the documentation and found no answer
- [ x ] I am running the latest version and the feature i am requesting is not implemented there

#### Describe the problem you are trying to solve
The Docker compose has some small ish flaws. Like mentioned it would be better if there was a docker image like described in [Issue #1382
](https://github.com/steve-community/steve/issues/1382). 

Some options to reconsidere are the exposing of the MariaDB port. Theoreticly not needed from what ive seen. 
And last but not least, the maria db dosnt use the volume for data persitency. 

#### Describe the solution you'd like

Add the Docker Image.
Add the volume mount.
Add a docker network for traffick from DB to the frontend
Remove the Port binding. 

#### Additional context
If i find time i try to propose a better compose file.

PR #1935, opened 2026-01-13
Update docker-compose.yml | Issue#1933

KjellWolf:
### **User description**
Making the DB volume persistent.
Adding options to use a .env file
Removing the DB Port mapping

Based on my Issue https://github.com/steve-community/steve/issues/1933

___

### **PR Type**
Enhancement

___

### **Description**
- Remove database port mapping for security

- Make database volume persistent with named volume

- Support environment variable configuration with defaults

- Improve database credentials security and flexibility

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["docker-compose.yml"] -->|Remove port mapping| B["Port 3306 hidden"]
  A -->|Add volume mount| C["db-data:/var/lib/mysql"]
  A -->|Add env variables| D["MYSQL_DATABASE<br/>MYSQL_USER<br/>MYSQL_PASSWORD"]
  D -->|Support defaults| E[".env file or defaults"]
```

<details><summary><h3>File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Configuration changes</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>docker-compose.yml</strong><dd><code>Persistent storage and environment variable configuration</code></dd></summary>
<hr>

docker-compose.yml

<ul><li>Removed port mapping (3306:3306) to prevent external database access<br> <li> Added named volume <code>db-data</code> for persistent database storage<br> <li> Replaced hardcoded credentials with environment variable references <br>using <code>${VAR:-default}</code> syntax<br> <li> Updated default password from <code>changeme</code> to <code>changeMe!</code> for better <br>security</ul>

</details>

  </td>
  <td><a href="https://github.com/steve-community/steve/pull/1935/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3">+5/-5</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr></tbody></table>

</details>

___

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,13 +11,13 @@ services:
     # MariaDB 10.4.
     image: mariadb:10.4.30
     restart: unless-stopped
-    ports:
-      - 3306:3306
     environment:
       MYSQL_RANDOM_ROOT_PASSWORD: "yes"
-      MYSQL_DATABASE: stevedb
-      MYSQL_USER: steve
-      MYSQL_PASSWORD: changeme
+      MYSQL_DATABASE: "${MYSQL_DATABASE:-stevedb}"
+      MYSQL_USER: "${MYSQL_USER:-steve}"
+      MYSQL_PASSWORD: "${MYSQL_PASSWORD:-changeMe!}"
+    volumes:
+      - db-data:/var/lib/mysql
     # https://mariadb.com/docs/server/server-management/automated-mariadb-deployment-and-administration/docker-and-mariadb/using-healthcheck-sh
     healthcheck:
       test: [ "CMD", "healthcheck.sh", "--connect", "--innodb_initialized" ]


docker-compose.yml
```diff
@@ -11,13 +11,13 @@ services:
     # MariaDB 10.4.
     image: mariadb:10.4.30
     restart: unless-stopped
-    ports:
-      - 3306:3306
```
goekay, line 15: hey there, thanks for the effort. 

this means that the database will not be accessible from localhost (let's say from a dev machine), right? it would be a little bit counter-productive IMO.

KjellWolf: Maybe there should be a Prod and Dev Compose file? or just  commented out?

Having unnessesary ports open for prod is dangerus. Sadly most people deploy without thinking.

goekay: hmm. to be frank, i never never considered this docker-compose (or any kubernetes-related additions) for prod. i think both of them are too context-dependent and should be tuned to the org's infra.

this docker-compose _for me_ is just a developer convenience shortcut to start steve easily in 1 step.

goekay: for example, _my_ definition/interpretation of prod would be some managed database by the cloud provider i am using, such that this mysql/mariadb in docker-compose is completely unnecessary. there are so many downsides IMO to run a database locally like this. but every org's prod choices, context and configuration is different. it is too nuanced to cover in this project.

KjellWolf: Well most people i know nowdays see docker and use docker.  Its just easy. Like you mentioned. And most People simly prefer containerized applications over bare metal installs

Especially if its not explicit stated that its Dev only. 

And IMO prod users should have the option for Docker too :D 

And maybe consider then "keep it simple" for the majority who just see and use and make it complex(er) for those who use it for Dev.

goekay: > And IMO prod users should have the option for Docker too :D

my suggestion was not to run the app on bare metal or without docker. i have nothing against docker. my argument was that the _dockerization_ of the app and the interplay with other components on your platform might look different from what is prescribed in the docker-compose. 

going back to your earlier argument, i guess we could have two docker-compose files: dev and prod. dev should be easy to use without modification. prod can be safer and more elaborate.

PR #1972, opened 2026-02-23
enforce inbound CALL messageId uniqueness per websocket session

goekay:
and use a compact hash-based store to do that

* track seen inbound CALL messageIds per WebSocket session
* reject duplicate CALL messageIds with ProtocolError
* wire deserializer validation through SessionContextStore lifecycle
* add dedicated IncomingMessageIdStore and integrate add/remove on session open/close
* memory/perf improvements for high-message stations: store Murmur3_128 fingerprints (asLong, 64-bit) instead of raw messageId strings. switch to fastutil LongOpenHashSet for lower per-entry overhead

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/ocpp/ws/IncomingMessageIdStore.java
+++ b/src/main/java/de/rwth/idsg/steve/ocpp/ws/IncomingMessageIdStore.java
@@ -20,6 +20,7 @@ package de.rwth.idsg.steve.ocpp.ws;
 
 import com.google.common.hash.HashFunction;
 import com.google.common.hash.Hashing;
+import de.rwth.idsg.steve.SteveException;
 import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
 import org.jetbrains.annotations.NotNull;
 import org.springframework.web.socket.WebSocketSession;
@@ -48,6 +49,13 @@ import java.util.concurrent.ConcurrentHashMap;
  * because calls are serialized by the outer per-chargeBox {@code Striped<Lock>} in
  * {@link SessionContextStoreImpl}.
  *
+ * <p> Capacity boundary:
+ * this store can enforce an optional per-session upper bound for unique incoming CALL messageIds.
+ * When the limit is exceeded, it throws an exception. This is intentional and causes the WebSocket session to be
+ * closed by the framework error handling path, avoiding unbounded growth for long-lived or misbehaving sessions.
+ * Regular charging stations should reconnect which will create a fresh session, effectively resetting duplicate-call-id
+ * tracking.
+ *
  * <p> Collision risk (birthday approximation, 64-bit stored key space):
  * p ~= n(n-1)/(2*2^64), where n is the number of messageIds in one session.
  * This is very small for practical values:
@@ -63,6 +71,16 @@ public class IncomingMessageIdStore {
     private static final HashFunction MURMUR3_128 = Hashing.murmur3_128();
 
     private final ConcurrentHashMap<String, LongOpenHashSet> lookupBySessionId = new ConcurrentHashMap<>();
+    private final Integer maxTrackedMessageIdsPerSession;
+
+    /**
+     * @param maxTrackedMessageIdsPerSession positive value enables bounded mode; {@code null} or non-positive value
+     *                                       disables the cap (unbounded mode).
+     */
+    IncomingMessageIdStore(Integer maxTrackedMessageIdsPerSession) {
+        boolean disable = (maxTrackedMessageIdsPerSession == null || maxTrackedMessageIdsPerSession <= 0);
+        this.maxTrackedMessageIdsPerSession = disable ? null : maxTrackedMessageIdsPerSession;
+    }
 
     void addSession(WebSocketSession session) {
         lookupBySessionId.computeIfAbsent(session.getId(), innerSession -> new LongOpenHashSet());
@@ -79,7 +97,19 @@ public class IncomingMessageIdStore {
             // session. it might imply some add/remove race or some other unexpected edge case.
             return null;
         }
+
         long fingerprint = MURMUR3_128.hashString(messageId, StandardCharsets.UTF_8).asLong();
-        return set.add(fingerprint);
+        boolean success = set.add(fingerprint);
+        if (!success) {
+            return false;
+        }
+
+        if (maxTrackedMessageIdsPerSession != null && set.size() > maxTrackedMessageIdsPerSession) {
+            throw new SteveException(
+                "Incoming CALL messageId limit of %s exceeded for session '%s'",
+                maxTrackedMessageIdsPerSession, session.getId()
+            );
+        }
+        return true;
     }
 }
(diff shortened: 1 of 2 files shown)


goekay:
review of several active OSS OCPP stacks and their public code/docs with the help of **_GPT 5.3 codex_**:

> ## OCPP `messageId` handling in OSS projects (full recap)
> 
> This summarizes findings for:
> - `mobilityhouse/ocpp` (Python)
> - `lorenzodonini/ocpp-go`
> - `mikuso/ocpp-rpc` (Node.js ecosystem style)
> - Scala OCPP lineage (`ShellRechargeSolutionsEU/ocpp` / `S44M/scala-ocpp`)
> - `citrineos/*` (CitrineOS)
> 
> It focuses on OCPP 1.6J §4.1.4 expectations for inbound `CALL.messageId` uniqueness per WebSocket session.
> 
> ---
> 
> ## Spec baseline (OCPP 1.6J §4.1.4)
> - `CALL`: sender must use a `messageId` not previously used on the same WS connection.
> - `CALLRESULT`/`CALLERROR`: `messageId` must match the original `CALL`.
> 
> ---
> 
> ## Executive summary
> - Most OSS stacks strongly implement **correlation** (`CALL` ↔ `CALLRESULT/CALLERROR`) and pending-call tracking.
> - Fewer stacks explicitly implement **strict inbound duplicate `CALL.messageId` rejection per session** as a dedicated mechanism.
> - Your implementation is stricter/more literal to §4.1.4 for inbound handling.
> 
> ---
> 
> ## Project-by-project recap
> 
> | Project | Typical `messageId` behavior | Inbound duplicate `CALL.messageId` guard (same WS session) |
> |---|---|---|
> | `mobilityhouse/ocpp` | Strong call/result correlation and protocol framing/validation | I found correlation enforcement (CALLRESULT/CALLERROR must match an outstanding call), but no clear evidence of a persistent “seen inbound CALL IDs” set. |
> | `ocpp-go` | Strong correlation and request lifecycle handling | I did not find evidence of explicit long-lived inbound duplicate-call-id rejection. |
> | `ocpp-rpc` style libs | RPC correlation, response matching, timeout/error flow | No documented evidence of per-session historical dedupe for inbound CALL IDs. |
> | Scala OCPP implementations | Typed protocol flow + request/response matching | I could not confirm a dedicated inbound duplicate-ID history set. |
> | CitrineOS | Router/broker architecture, context/correlation (`correlationId`) through processing pipeline | I did not find clear evidence of strict inbound duplicate-call-id tracking. |
> 
> ---
> 
> ## How this compares to your implementation
> 
> - Your implementation now goes beyond the typical OSS correlation-first approach by enforcing **inbound per-session `CALL.messageId` uniqueness** explicitly.
> - It uses **compact fingerprint storage** (`murmur3_128().asLong()` + `LongOpenHashSet`) rather than raw `String` IDs, reducing memory pressure substantially.
> - You added a **session-level safety boundary**:
>   - default cap: **1,000,000 unique inbound CALL IDs per session**
>   - cap is **optional** (can be disabled for unbounded mode).
> - On overflow, the store throws `SteveException`; in your current Spring WebSocket stack this propagates through the handler path and is handled by the framework’s exception decorator, which closes the socket with server error status.

goekay:
_How much memory will be consumed by LongOpenHashSet with 1 million entries per session?_

> Roughly **~16.8 MB per session** for the set itself.
> 
> Quick breakdown (default fastutil load factor `0.75`):
> - Needed slots for 1,000,000 entries: `ceil(1,000,000 / 0.75) = 1,333,334`
> - Next power-of-two table size: `2,097,152`
> - `LongOpenHashSet` stores keys in a `long[]` of size `n + 1`:
>   - `(2,097,152 + 1) * 8 bytes = 16,777,224 bytes` (~16.0 MiB)
> 
> With object/header/alignment overhead, practical total is typically around **16.8–17.5 MB** per session.

This means that the memory consumption of the app will correlate (and grow) with the number of connected stations. 100 connected stations (assuming 1 session per station) means _**up to**_ 1.8 GB memory just for messageId bookkeeping, if all these connections ever would reach 1 million message limit. My pragmatic assumption is that stations will disconnect & reconnect due to other reasons, such that this cap will hopefully remain a theoretical upper boundary. We will need to monitor the behavior closely and make adjustments when/if necessary.

goekay:
@juherr wdyt?

juherr:
@goekay It looks solid.
One risk to keep in mind is that a station may retry the same message multiple times. You'll only know how significant that is once this runs into production.

juherr:
Regarding data storage, would it make sense to define a storage interface with an in-memory implementation by default and optionally provide a KV-backed implementation for deployments with a large number of stations?

This would keep the simple setup lightweight while allowing more scalable environments to plug in a more suitable persistence layer without changing the core logic.

goekay:
> One risk to keep in mind is that a station may retry the same message multiple times.

actually, i was looking at ocpp 1.6, 2.0.1 (Edition 3 FINAL, 2024-05-06) and 2.1 (Edition 1, 2025-01-23) for this particular case. 

ocpp 1.6 is simple: every CALL on each session needs to use different messageId. this also covers retries implicitly.

ocpp 2.0.1 is explicit about this behaviour:

Screenshot 2026-02-23 at 16 49 59

ocpp 2.1 takes a step back from this strictness, and has language about messageId reuse for message retry:

Screenshot 2026-02-23 at 16 44 51

so... in short, this impl should be compliant with 1.6. if we face issues, we can always reduce strictness in a future step.

> Regarding data storage, would it make sense to define a storage interface with an in-memory implementation by default and optionally provide a KV-backed implementation for deployments with a large number of stations?

this would be such a simple/cheap/fast change: extracting an interface from IncomingMessageIdStore and providing skeleton impl for KV. i rather would do it if it is needed (either by us/Powerfill, or someone requests this). so, i am open for this but whenever this requirement actually comes.

juherr:
If retried messages are already considered acceptable in 2.x, why not apply the same tolerance to 1.6? 
Instead of rejecting them, we could accept the message but flag it as suspected spam from a known station.
This introduces no functional risk, while providing valuable operational feedback and helping to pinpoint problematic devices.

I implemented a similar approach at the WebSocket connection level to identify stations with unstable network behavior. 
We could follow the same philosophy here: no hard failure, but actionable insight for station maintenance and monitoring.

goekay:
> If retried messages are already considered acceptable in 2.x, why not apply the same tolerance to 1.6?

multiple reasons: 

1. well, first of all, it would not be correct according to 1.6 spec. 
2. the impl needs to make sure that reuse of messageId is indeed a retry for the same exact message. for this, looking at messageId is not enough. bigger context is needed: we would need to look at the message body as well. so, suddenly we have to keep track of that as well (maybe just the footprint/hash to reduce the overhead). otherwise, two different message bodies can share the messageId which is not the intent. i dont want to increase the scope and implementation complexity for now. 
3. ocpp 2.x also signals that the messageIds should be unique during the existence of station, period. there is no language for _unique within "a websocket session" of station_. this actually implies a durable storage for all incoming messages, and validating against all messages. so, whenever we do ocpp 2.x it will be probably a complete overhaul of this feature. therefore, i can imagine that it will be a parallel implementation. applying this tolerance now is not helping us in the future.

PR #2013, opened 2026-04-17
2012 Add bidirectional binding: Change Configuration and Database CRUD for OCPP security

### LINKED ISSUE #2012: Add bidirectional binding: Change Configuration and Database CRUD for OCPP security
Currently, security profile and basic auth password can be changed at two places:
1. charge point update CRUD
2. OCPP operation: Change Configuration with keys `AuthorizationKey` and `SecurityProfile`

These two are disconnected flows right now: Updating the database (1) does not trigger a Change Configuration to inform the station. Or, using Change Configuration to update the two fields in station does not update the database (2).

It would be nice (and less confusing) to have bidirectional binding/integration between the two to streamline operations.

goekay:
https://github.com/steve-community/steve/issues/2012


src/main/java/de/rwth/idsg/steve/service/ChargePointConfigUpdater.java
```diff
@@ -0,0 +1,135 @@
...
+            if (status == null) {
+                throw new SteveException("Missing ChangeConfiguration response for key '%s' from chargeBoxId '%s'".formatted(configKey, chargeBoxId));
+            }
+
+            if (status != ConfigurationStatus.ACCEPTED) {
+                throw new SteveException("Charge point '%s' rejected ChangeConfiguration key '%s' with status '%s'".formatted(chargeBoxId, configKey, status.value()));
+            }
+        } catch (SteveException.BadRequest e) {
+            log.warn("Skipping ChangeConfiguration for key={} on offline/unreachable chargeBoxId={}", configKey, chargeBoxId);
+        } catch (SteveException e) {
```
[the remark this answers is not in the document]
goekay: this is okay and by design. we should be able to change profile/pwd of stations that are offline now. use case: i added a new station to steve, did not even connect it to steve yet, want to change the profile/pwd. so, there is a difference between a) an existing station is temporarily disconnected and b) this station did not even connect yet. instead of surfacing the error to user which can be annoying, we just log a warning for the operator to notice.

src/main/java/de/rwth/idsg/steve/service/ChargePointService.java
```diff
@@ -118,8 +136,29 @@ public int addChargePoint(ChargePointForm form) {
...
+        // but station will not know about them. whenever this happens, it does not happen silently though. the function
+        // calls below will throw exceptions which we bubble up to the user. then, it becomes the operational concern
+        // of the user to follow up.
+        entry.ifPresent(fromDatabase -> {
+            if (fromDatabase.securityProfile() != newSecurityProfile) {
+                applicationEventPublisher.publishEvent(new OcppStationSecurityProfileChanged(chargeBoxId, newSecurityProfile));
+            }
+            if (StringUtils.isNotBlank(newRawPassword) && !passwordEncoder.matches(newRawPassword, fromDatabase.hashedAuthPassword())) {
+                applicationEventPublisher.publishEvent(new OcppStationSecurityBasicAuthChanged(chargeBoxId, newRawPassword));
+            }
```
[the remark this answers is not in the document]
goekay: false positive.

java docs from `org.springframework.security.crypto.password.PasswordEncoder`:

> Never true if either rawPassword or encodedPassword is null or an empty String.

so, `PasswordEncoder.matches("foo", null)` and `PasswordEncoder.matches("foo", "")` both return `false`, which we negate in if-condition upon which we would go inside the if-block to publish the station-update event.

src/main/java/de/rwth/idsg/steve/service/ChargePointService.java
```diff
@@ -147,10 +147,15 @@ public void updateChargePoint(ChargePointForm form) {
...
         // if securityProfile or authPassword changed, try to change these at the station as well.
-        // we do this on best-effort principle: if the call fails or the station rejects, we will have new values in DB
+        //
+        // 1) we do this on best-effort principle: if the call fails or station rejects, we will have new values in DB
         // but station will not know about them. whenever this happens, it does not happen silently though. the function
         // calls below will throw exceptions which we bubble up to the user. then, it becomes the operational concern
         // of the user to follow up.
+        // 2) it is crucial for registered EventListeners NOT to be async such that exceptions can be visible to user.
+        // 3) these two keys are only applicable to OCPP 1.6J stations with improved security. any other station will
+        // reject them. we do not filter/validate here with the assumption that the user knows what she is doing.
```
juherr:
```diff
-the user knows what she is doing
+the user knows what they are doing
```

This wording feels a bit unusual and might be worth revisiting.

---- 

Would it make sense to inject ChargePointConfigUpdater here instead of relying on the event bus?

Aside from `var registration = chargePointService.getRegistrationDirect(chargeBoxId);` are there other circular dependencies to consider?
If registration is the main constraint, perhaps it could be included in the event payload to help break the cycle.

goekay: > This wording feels a bit unusual and might be worth revisiting.

done. thanks.

> Would it make sense to inject ChargePointConfigUpdater here instead of relying on the event bus?

i did multiple attempts due to circular dependencies. tried to break them cleanly. was not happy with any. also did not want to overcomplicate by slicing/splitting like crazy. this is the acceptable final result considering pros/cons.

> If registration is the main constraint, perhaps it could be included in the event payload to help break the cycle.

the second-order effects (i.e. the wave this would create) is not worth it IMO. but also: we fetch registration lazily for only ocpp version=1.6. i would not want to do it generally. so, building this conditional solution is not worth it.

juherr: I would probably take a slightly different approach here.

Since OCPP 1.6+ is the dominant use case, an additional fetch for the remaining cases would likely be acceptable. In exchange, it could simplify the design and reduce the need for some of the current trade-offs.

PR #2039, opened 2026-05-06
get and store all configs of the station (instead of just CpoName)

goekay:
we currently GetConfig from stations (after station connecting) for the CpoName key and store it in database. we can generalize this flow to get all config keys/values from stations and store them in database.

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/service/ChargePointService.java
+++ b/src/main/java/de/rwth/idsg/steve/service/ChargePointService.java
@@ -67,8 +67,6 @@ import java.util.concurrent.locks.Lock;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import static de.rwth.idsg.steve.web.dto.ocpp.ConfigurationKeyEnum.AuthorizationKey;
-
 /**
  * @author Sevket Goekay <sevketgokay@gmail.com>
  * @since 29.10.2025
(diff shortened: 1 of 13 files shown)


src/main/java/de/rwth/idsg/steve/service/ChargePointService.java
```diff
@@ -126,6 +120,26 @@ public void updateSecurityProfile(String chargeBoxId, String securityProfile) {
...
+        try {
+            chargePointRepository.updateOcppConfiguration(chargeBoxId, jsonNode);
+        } catch (Exception e) {
+            log.error("Failed during updateOcppConfiguration, because: {}", e.getMessage(), e);
+        }
+    }
+
+    public void updateOcppConfigurationAfterChange(String chargeBoxId, String key, String value) {
+        if (StringUtils.isEmpty(key) || AuthorizationKey.name().equals(key)) {
+            return;  // AuthorizationKey is write-only, we should not store it like that
```
juherr: Be careful: depending on the station model and/or firmware version, it could exist more keys that are write-only.
As an example, see 

Image

https://knowledge.alfen.com/download/attachments/243466257/ACE%20NG9%20Backoffice%20Configuration%20keys%206.6.0_v2.pdf

goekay: hmm. i dont know why these particular keys should be write only. they dont seem to be sensitive (as the plain password is).

but i get your point. i guess the only proper way is to make a GetConfiguration request after the change, hoping/assuming that the station will respond with the updated value(s) and excluding write-only keys.

juherr: From my experience, some Alfen stations may return a different value from the one pushed by the CSMS for certain configuration keys, even for some read/write keys.

Here the best/worst example:
image
(passwords will be replaced by `***`)
https://knowledge.alfen.com/space/IN/243466257/Release+notes+-+Firmware+updates+NG-Platform+(NG-910+%2F+NG-920)?attachment=https://knowledge.alfen.com/download/attachments/243466257/ACE%2520NG9%2520Backoffice%2520Configuration%2520Keys.pdf&type=application/pdf&filename=ACE+NG9+Backoffice+Configuration+Keys.pdf

If you always store the value returned by the station as the source of truth, the original value pushed by the CSMS may be lost. This could be problematic in some cases.

PR #2090, opened 2026-07-27
add OCA certification note for Powerfill

goekay:


juherr:
Congratulations @goekay 👏

PR #2093, opened 2026-07-28
Powerfill donates OCTT workarounds

goekay:
## Context

This PR contributes changes from the [OCA certification of Powerfill’s OCPP 1.6 CSMS](https://powerfill.io/blog/powerfill-oca-certified-ocpp-1-6-core-smart-charging-advanced-security).

These changes do not correct OCPP violations in SteVe. The previous behavior was standards-compliant. During certification, however, several OCTT scenarios did not accept or support that behavior. The resulting accommodations are isolated behind an explicit compatibility setting so that SteVe’s default OCPP behavior remains unchanged.

## Changes

- Add the opt-in `steve.ocpp.octt-quirks-enabled` flag.
- Normalize certificate and certificate-chain inputs to LF line endings. Since this is a general-enough normalization, it is not behind the compatibility flag.

When OCTT compatibility is enabled, SteVe additionally:

- uses `SHA256withRSA` instead of the OCPP-prescribed RSA-PSS algorithm for RSA certificate signatures;
- skips certificate subject validation because OCTT does not provide the reference information required;
- suppresses automatic `GetConfiguration` requests after a WebSocket connection and after `ChangeConfiguration`, because OCTT does not expect these otherwise valid sequences of events.

## Compatibility and standards behavior

`octt-quirks-enabled` defaults to `false`.

The compatibility mode intentionally alters or relaxes specific standards-aligned behavior solely to support OCTT-compatibility. It should not be enabled in regular production deployments.

These changes originate from the Powerfill certification effort. They do not imply that standalone SteVe builds or deployments are themselves OCA-certified.

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/config/SteveProperties.java
+++ b/src/main/java/de/rwth/idsg/steve/config/SteveProperties.java
@@ -68,6 +68,7 @@ public class SteveProperties {
         boolean autoRegisterUnknownStations;
         String chargeBoxIdValidationRegex;
         Protocols enabledProtocols;
+        boolean octtQuirksEnabled;
         Security security = new Security();
 
         @Data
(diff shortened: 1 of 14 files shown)


juherr:
@shiv3 fyi

PR #2119, opened 2026-08-22
Switch to use maintained nv-i18n package

AlanGraham:
The original nv-i18n is no longer being maintained. Our company app relies on it, and needed newly created entries. A colleague and I have forked and published a new version of it. Due to our reliance on it we will be continuing to maintain it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -615,11 +615,11 @@
         </dependency>
 
         <!-- Codes according to ISO standards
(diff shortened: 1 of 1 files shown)


goekay:
thank you.

background: https://github.com/TakahikoKawasaki/nv-i18n/issues/96#issuecomment-5221227355

PR #2122, opened 2026-08-26
perf(ocpp): send WebSocket pings from virtual threads

juherr:
Rebased on master now that #2121 is merged, and reduced to what @goekay asked for: a virtual-thread
scheduler and the wiring that hands it to the session store. 2 files, +17/-1.

## Why

The pings run on the shared `TaskScheduler`, a pool of ten threads that every other periodic job of SteVe
also uses. A station whose send blocks holds one of them for up to the ten second send limit. Now that
the interval is configurable (#3), an operator who lowers it to work around a proxy timeout also
multiplies the pressure on that pool.

Holding N schedules is not the problem: `ThreadPoolTaskScheduler` is backed by a
`ScheduledThreadPoolExecutor`, whose delay queue is a binary heap, so ten thousand due dates cost
`O(log N)` per firing and the thread sleeps in between. Where the ping *runs* is the problem.

## What

`webSocketPingScheduler`, a `SimpleAsyncTaskScheduler` with virtual threads enabled, sits next to
`taskScheduler()` in `BeanConfiguration`, and `SessionContextStoreHolder` asks for it by qualifier. One
thread keeps the due dates; each ping runs on its own virtual thread. `PingTask`, `SessionContextStoreImpl`
and `SessionContext` are untouched.

One consequence worth knowing: `SimpleAsyncTaskScheduler` dispatches every execution to a new thread
rather than running it on the scheduler thread, so `scheduleAtFixedRate` no longer guarantees that two
executions of the same task never overlap. A session whose send blocks for longer than the ping interval
can have more than one ping in flight. `ConcurrentWebSocketSessionDecorator` already bounds that case: it
closes a session whose send exceeds its 10s limit or whose buffer fills up.

## Tests

No new test — what is left is a bean definition, and the behaviour under it is Spring's. It is covered end
to end by `Issue3IT`, where the ping now goes out on `SteVe-WebSocketPing-2` instead of
`SteVe-TaskScheduler-*`.

---
*Drafted by Claude under @juherr's supervision.*

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/config/BeanConfiguration.java
+++ b/src/main/java/de/rwth/idsg/steve/config/BeanConfiguration.java
@@ -40,6 +40,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Primary;
 import org.springframework.scheduling.annotation.EnableAsync;
 import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.scheduling.concurrent.SimpleAsyncTaskScheduler;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
 import org.springframework.web.servlet.config.annotation.ResourceHandlerRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
@@ -167,6 +168,20 @@ public class BeanConfiguration implements WebMvcConfigurer {
         scheduler.setWaitForTasksToCompleteOnShutdown(true);
         scheduler.setAwaitTerminationSeconds(30);
         scheduler.initialize();
+
+        return scheduler;
+    }
+
+    /**
+     * The WebSocket pings are scheduled here rather than on the scheduler above, which they would share
+     * with every other periodic job of SteVe. This one runs each ping on its own virtual thread, so a
+     * station whose send blocks costs a virtual thread instead of a tenth of that pool.
+     */
+    @Bean
+    public SimpleAsyncTaskScheduler webSocketPingScheduler() {
+        SimpleAsyncTaskScheduler scheduler = new SimpleAsyncTaskScheduler();
+        scheduler.setThreadNamePrefix("SteVe-WebSocketPing-");
+        scheduler.setVirtualThreads(true);
 
         return scheduler;
     }
--- a/src/main/java/de/rwth/idsg/steve/ocpp/ws/SessionContextStoreHolder.java
+++ b/src/main/java/de/rwth/idsg/steve/ocpp/ws/SessionContextStoreHolder.java
@@ -21,6 +21,7 @@ package de.rwth.idsg.steve.ocpp.ws;
 import de.rwth.idsg.steve.config.SteveProperties;
 import de.rwth.idsg.steve.ocpp.OcppVersion;
 import de.rwth.idsg.steve.ocpp.ws.custom.WsSessionSelectStrategy;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.scheduling.TaskScheduler;
 import org.springframework.stereotype.Component;
 
@@ -42,7 +43,7 @@ public class SessionContextStoreHolder {
     private final FutureResponseContextStore futureResponseContextStore;
 
     public SessionContextStoreHolder(SteveProperties steveProperties,
-                                     TaskScheduler taskScheduler,
+                                     @Qualifier("webSocketPingScheduler") TaskScheduler taskScheduler,
                                      FutureResponseContextStore futureResponseContextStore) {
         wsSessionSelectStrategy = steveProperties.getOcpp().getWsSessionSelectStrategy();
         pingInterval = steveProperties.getOcpp().getWsPingInterval();


goekay:
hey @juherr this is a nice improvement. since the other PR is merged, i would like to merge this if you could rebase this and keep only what is relevant/essential for the topic.

src/main/java/de/rwth/idsg/steve/ocpp/ws/PingTask.java
```diff
@@ -33,26 +35,72 @@
...
+        this.executor = executor;
+    }
 
+    /**
+     * Runs on the scheduler thread and only hands the ping over.
+     *
+     * {@code scheduleAtFixedRate} never runs two executions of the same task at once. Since we return
+     * before the send is done, keeping that promise is now our job: a session whose send blocks for
+     * longer than the ping interval must not pile up pings.
+     */
```
goekay: for the sake of simplification: why dont we just execute all pings in a dedicated SimpleAsyncTaskExecutor spring bean that has virtual threads enabled? isn't this handover unnecessarily convoluted?

juherr: Done in 21fe107.

`webSocketPingExecutor` is now a `SimpleAsyncTaskExecutor` bean with virtual threads enabled, declared
next to `taskScheduler()` in `BeanConfiguration` and injected through `SessionContextStoreHolder` →
`SessionContextStoreImpl` → `PingTask`. `PingTask` loses its static executor and its second constructor,
and the bean is `AutoCloseable`, so the context closes it at shutdown — which the static never was.

What is left of the hand-over is two lines in `run()` plus the `inFlight` flag, and the flag is the part
that is not plumbing: `scheduleAtFixedRate` guarantees that two executions of the same task never
overlap, and that guarantee is gone the moment `run()` returns before the send is done. Without it, a
station whose send blocks for longer than the ping interval gets one more ping queued behind it per
firing — the regression Qodo caught on #2121. Happy to drop it if you would rather rely on
`ConcurrentWebSocketSessionDecorator` closing such a session once it hits its 10s send limit.

I also dropped `testSendRunsOnAVirtualThread`: with the executor coming from the context, "the ping runs
on a virtual thread" is a property of the bean definition rather than of `PingTask`, so the test was only
asserting Spring's behaviour. `PingTaskTest` is now four synchronous, deterministic tests.

---
*Drafted by Claude under @juherr's supervision.*

juherr:
> **An async test can outlive its worker** — `testSendRunsOnAVirtualThread` waits only until `sendMessage`
> counts down the latch and never waits for the captured `sendingThread` to terminate.

Valid as written, and moot as of 21fe107: that test is gone. Following @goekay's review, the virtual-thread
executor is now a Spring bean (`webSocketPingExecutor`) rather than a static inside `PingTask`, so "the
ping runs on a virtual thread" is a property of the bean definition, not something `PingTask` can be
asked about. `PingTaskTest` now drives a synchronous executor: four deterministic tests, no latch, no
worker left to join.

---
*Drafted by Claude under @juherr's supervision.*

goekay:
the regular taskscheduler still does the time scheduling, and the execution there hands over to the virtual thread executor. this hand-over should not be necessary. the only necessary change is creating a virtual-enabled scheduler, and passing it down the path. there is no need to touch lower levels like PingTask etc

src/main/java/de/rwth/idsg/steve/ocpp/ws/PingTask.java
juherr: You are right, and the version above was two layers too deep. Reworked in 5a352e0: the only change is a
virtual-thread scheduler and the wiring that hands it to the session store.

`webSocketPingScheduler` is a `SimpleAsyncTaskScheduler` with `setVirtualThreads(true)`, next to
`taskScheduler()` in `BeanConfiguration`, and `SessionContextStoreHolder` asks for it by qualifier.
`PingTask`, `SessionContextStoreImpl` and `SessionContextStoreTest` are back to exactly what is on
master; `PingTaskTest` is gone. The whole PR is now 2 files, +17/-1.

One consequence to be aware of, since it is the reason the previous version carried an in-flight flag:
`SimpleAsyncTaskScheduler` dispatches every execution to a new thread instead of running it on the
scheduler thread, so `scheduleAtFixedRate` no longer guarantees that two executions of the same task
never overlap. A session whose send blocks for longer than the ping interval can therefore have more
than one ping in flight. `ConcurrentWebSocketSessionDecorator` already bounds that case — it closes a
session whose send exceeds its 10s limit or whose buffer fills up — so I left it at that rather than
reintroducing a guard.

No new test: what is left is a bean definition, and the behaviour under it is Spring's. It is covered
end to end by `Issue3IT`, where the ping now goes out on `SteVe-WebSocketPing-2` instead of
`SteVe-TaskScheduler-*`.

---
*Drafted by Claude under @juherr's supervision.*

goekay: thanks, that should be it.

PR #2125, opened 2026-08-27
migrate to in-memory message queue

### LINKED ISSUE #2104: Introduce a message queue for OCPP message processing
SteVe currently consists of a single Java application and a relational database. Incoming OCPP messages are received, processed, persisted, and answered directly inside the Java application. This tightly couples the OCPP transport layer to message processing.

### Goal

Introduce a message queue between OCPP transport and OCPP processing.

The Java application will remain a single deployable application. This issue does not introduce microservices or split the existing application (maybe later).

The application will contain:
- A producer that publishes incoming raw OCPP messages.
- An inbound listener that consumes and processes those messages.
- A producer that publishes OCPP responses.
- An outbound listener that consumes responses and sends them to the appropriate charging station.

### New architecture draft

Image

### Open topics (TBD)

- Message envelope: What particular fields and metadata in addition to JSON/SOAP payload.
- The actual message queue solution: RabbitMQ or something more modern and lightweight?
- Both for SOAP and JSON, or safe to ignore SOAP at last?

goekay:
Message flow is now:
- all incoming: WebSocketEndpoint → inbound queue  → IncomingPipeline
- outgoing responses: IncomingPipeline  → outbound queue → OutgoingPipeline
- outgoing calls: ChargePointServiceJsonInvoker → outbound-call queue → OutgoingPipeline

### DIFF
--- a/src/main/java/de/rwth/idsg/steve/ocpp/ws/data/CommunicationContext.java
+++ b/src/main/java/de/rwth/idsg/steve/ocpp/ws/data/CommunicationContext.java
@@ -22,6 +22,7 @@ import de.rwth.idsg.steve.ocpp.OcppProtocol;
 import lombok.Getter;
 import org.jetbrains.annotations.Nullable;
 
+import java.time.Instant;
 import java.util.UUID;
 
 /**
@@ -40,7 +41,7 @@ public class CommunicationContext {
         String chargeBoxId,
         OcppProtocol protocol,
         String webSocketSessionId,
-        long arrivedAtMillis,
+        Instant receivedAt,
 
         String ocppPayload // full and raw JSON array payload
     ) {
--- a/src/main/java/de/rwth/idsg/steve/ocpp/ws/data/CommunicationContext.java
+++ b/src/main/java/de/rwth/idsg/steve/ocpp/ws/data/CommunicationContext.java
@@ -40,6 +40,7 @@ public class CommunicationContext {
         String chargeBoxId,
         OcppProtocol protocol,
         String webSocketSessionId,
+        long arrivedAtMillis,
 
         String ocppPayload // full and raw JSON array payload
     ) {
--- a/src/main/java/de/rwth/idsg/steve/ocpp/ws/pipeline/IncomingPipeline.java
+++ b/src/main/java/de/rwth/idsg/steve/ocpp/ws/pipeline/IncomingPipeline.java
@@ -24,6 +24,7 @@ import de.rwth.idsg.steve.config.MessagingConfiguration;
 import de.rwth.idsg.steve.messaging.Messaging;
 import de.rwth.idsg.steve.ocpp.OcppVersion;
 import de.rwth.idsg.steve.ocpp.ws.data.CommunicationContext;
+import de.rwth.idsg.steve.ocpp.ws.data.OcppJsonResponse;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.integration.annotation.ServiceActivator;
@@ -74,8 +75,7 @@ public class IncomingPipeline implements Messaging.In.Consumer {
 
         } catch (CommunicationContext.JsonCallParseException e) {
             var parseError = e.getParseError();
-            var parseErrorStr = serializer.accept(parseError);
-            sendOut(message, parseErrorStr);
+            sendOut(message, parseError);
             return;
 
         } catch (CommunicationContext.JsonResponseInvalidException e) {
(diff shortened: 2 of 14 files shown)


goekay:
@juherr wdyt?

src/main/java/de/rwth/idsg/steve/ocpp/ws/data/CommunicationContext.java
```diff
@@ -40,6 +40,7 @@ public record In(
         String chargeBoxId,
         OcppProtocol protocol,
         String webSocketSessionId,
+        long arrivedAtMillis,
```
juherr:
```suggestion
        Instant receivedAt,
```

Could this be an `Instant receivedAt` directly?

`receivedAt` seems a little clearer than `arrivedAt`, and keeping it as an `Instant` avoids carrying an implicit millisecond unit through the application only to convert it back later.

src/main/java/de/rwth/idsg/steve/ocpp/ws/pipeline/IncomingPipeline.java
```diff
@@ -98,7 +106,13 @@ private void processCall(CommunicationContext.InCall data) {
 
         var response = handler.accept(data);
         var responseStr = serializer.accept(response);
-        outgoingPipeline.accept(CommunicationContext.outFrom(data.in(), responseStr));
+        sendOut(message, responseStr);
+    }
+
+    private void sendOut(Message<CommunicationContext.In> inMsg, String ocppPayload) {
```
juherr:
```suggestion
    private void sendOut(Message<CommunicationContext.In> inMsg, ResponseType response) {
```

Would it make sense to move the serialization into `sendOut`?

That would let the pipeline keep the typed OCPP response until the actual outgoing-message boundary instead of converting it to a `String` earlier than necessary.

And could we introduce a dedicated `sendOutError(...)` here, so error serialization stays encapsulated and `sendOut(...)` can remain focused on typed OCPP responses?

juherr (commented):
I like the updated approach.

I am usually cautious about leaking too much Spring-specific code into the domain/application layers, but in this case I think Spring Integration is a good compromise. It gives us a proper messaging abstraction today while keeping a realistic path toward RabbitMQ later without introducing a custom abstraction just for the sake of hiding Spring.

The main point I would still address is ordering: **messages from the same charging station must be processed in receive order**.

With the current 4-thread `ExecutorChannel`, two consecutive messages from the same station can be processed concurrently and complete out of order.

Ideally, I would partition by `chargeBoxId`, because ordering is really a station-level concern. `PartitionedChannel` seems to fit that model well: parallel processing across stations, sequential processing within one partition.

There is one edge case worth thinking about though: SteVe technically allows multiple WebSocket sessions for the same charge point. In practice, multiple sessions are usually related to reconnects, so `chargeBoxId` still feels like the right partition key, but we should make sure concurrent/overlapping sessions cannot create unexpected ordering issues.

The timeout changes and the other Qodo fixes look good to me. I would mainly clarify and enforce this ordering guarantee before merging.

src/main/java/de/rwth/idsg/steve/ocpp/ws/pipeline/IncomingPipeline.java
goekay: > Would it make sense to move the serialization into sendOut?

sure. missed it.

> And could we introduce a dedicated ...

not necessary, because both result and error types are of OcppJsonResponse type and we can call sendOut(..) based on them.

src/main/java/de/rwth/idsg/steve/ocpp/ws/data/CommunicationContext.java
goekay: sure.

goekay:
@juherr i have multiple opinions/arguments about https://github.com/steve-community/steve/pull/2125#pullrequestreview-5043149444

1. i had a working prototype for PartitionedChannel (which was easy to add btw) that i discarded, because i want to keep things simple.
2. if you think about it, having 4 consumer threads is _less parallelism_ (or: _more serialization_) compared to what we had before. websocket messages were picked up and processed by the general thread pool of the server. the same pool that handles all HTTP work (min 8 and max 200 threads for spring boot jetty as default). and we did not experience any downside of this parallelism so far (also not with [powerfill](https://powerfill.io/)).
3. about "messages from the same charging station must be processed in receive order.": 
i went back to spec multiple times over the years for this particular topic. actually, no. the spec puts this responsibility on sender and not receiver:  

Screenshot 2026-08-28 at 09 53 07

because of these arguments, i think current impl is okay. you can challenge my arguments though :) 

edit: actually, by the same definition of this synchronicity we should serialize outgoing channel per station. but that alone would not be enough. a CALL message N in queue should only be sent whenever a response to a previous call N-1 arrived. so, we actually need some connection between outbound and inbound queue to really fullfill this aspect of spec. we did not have it this far either.

juherr:
Thanks, I learned something from your explanation about this part of the spec.

I agree that `PartitionedChannel` is probably not necessary here.

That said, I still think SteVe should ideally detect when a station violates the synchronicity rule. I see the CSMS as a buffer between imperfect charging stations and external systems such as APIs or OCPI, and part of its value is reducing the impact of inconsistent station behavior.

Even a warning or metric could already be useful for diagnostics.

And yes, enforcing the rule on SteVe's outbound CALLs sounds like a separate topic.

PR #2127, opened 2026-08-28
Swagger / API Doc using contex.path

fnkbsi:
If SteVe uses the contex.path in the configuration, the try-out functionality of the swagger RestApi-Doc-Site does not work. 
This PR fix this by adding the context.path.


goekay:
hey @fnkbsi thank you very much for this catch.

a steve-specific field in config file should not be necessary, since there is already a spring boot field for it: `server.servlet.context-path` and we already set it. you could just inject `org.springframework.boot.web.server.autoconfigure.ServerProperties` and read the path from this object to set the api docs relative path.

otherwise, LGTM.

PR #2138, opened 2026-09-18
Connect docker volume to database service

### LINKED ISSUE #2136: Docker database volume not used
Here the volume is declared

https://github.com/steve-community/steve/blob/9e6108c2c66bb1ee574f173ae310fe6d99b7b750/docker-compose.yml#L3-L5

but then it's not applied to the database

https://github.com/steve-community/steve/blob/9e6108c2c66bb1ee574f173ae310fe6d99b7b750/docker-compose.yml#L9-L19

it should be something like this

```yaml
  db:
    image: mariadb:10.11.16
    restart: unless-stopped
    ports:
      - 3306:3306
    environment:
      MYSQL_RANDOM_ROOT_PASSWORD: "yes"
      TZ: +00:00
      MYSQL_DATABASE: stevedb
      MYSQL_USER: steve
      MYSQL_PASSWORD: changeme
    volumes:
      - db-data:/var/lib/mysql
```

LivingWithHippos:
close #2136


goekay:
thank you!

